PR #8, opened 2021-08-19
feat: Configurable pick realm algorithm

pablitar:
This is the implementation of what was discussed in this RFC: https://docs.google.com/document/d/1uLj0ZGT_41SotfVVQIzUY4WtGef1yZ2x1t9M_mcjtDc/edit

This revamps the algorithm to consider close peers, and also to make it configurable through feature flags. Eventually, if we know a particular event is coming or something like that, we can change the algorithm to be mostly load balancing. Or we could tune the parameters when catalysts are updated. Also, it should be quite easy to add new links to the algorithm chain.

By default it uses the configurations in the defaults file: https://github.com/decentraland/kernel/pull/8/files#diff-b03faa0f792b66f82dba9fb83a23e9a13a135d8909bac054d968668214603b26


menduz (approved):

packages/shared/dao/pick-realm-algorithm/allPeers.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    const linearUsersScore = (users: number) => baseScore + users
+
+    if (max) {
+      // We try to fill all realms until around the percentage provided
+      if (count >= fillTargetPercentage * max) {
+        // If this is the case, we are in the "downward" phase of the score
+        // We calculate a segment joining the fillTargetPercentage% of users with baseScore at discourageFillTargetPercentage% maxUsers
+        // In that way, when we reach discourageFillTargetPercentage% maxUsers, realms that have at least one user start to get prioritized
+        const segment = { a: { x: fillTargetPercentage * max, y: linearUsersScore(fillTargetPercentage * max) }, b: { x: discourageFillTargetPercentage * max, y: baseScore } }
```
jmoguilevsky, line 21: can we rename `a` and `b` to `startingPoint` `endingPoint` or something a bit more descriptive?

pbosio (approved):

packages/shared/dao/pick-realm-algorithm/allPeers.ts
```diff
@@ -0,0 +1,41 @@
...
+  function usersScore(candidate: Candidate) {
+    const count = usersCount(candidate)
+    const max = maxUsers(candidate)
+
+    // We prefer realms that have users. Those will have at least baseScore
+    if (count === 0) return 0
+
+    const linearUsersScore = (users: number) => baseScore + users
+
+    if (max) {
```
agusaldasoro, line 15: why can `max` be undefined?

packages/shared/dao/index.ts
```diff
@@ -204,73 +190,42 @@ export async function fetchCatalystStatuses(nodes: { domain: string }[]) {
...
-  if (sorted.length === 0 && candidates.length > 0) {
-    throw new Error('No available realm found!')
+  if (!config || config.length === 0) {
+    config = defaultChainConfig
   }
 
-  return candidateToRealm(sorted[0])
+  const algorithm = createAlgorithm(config)
+
+  return candidateToRealm(algorithm.pickCandidate(candidates, currentUserParcel))
```
agusaldasoro, line 228: Is the parcel always present?

agusaldasoro (approved):

packages/shared/dao/pick-realm-algorithm/allPeers.ts
pablitar: Archipelago servers currently don't have maxUsers... That'll probably change, but will also probably change to something so big that this will be not so relevant... We'll see what happens then.

packages/shared/dao/pick-realm-algorithm/allPeers.ts
pablitar: done

packages/shared/dao/index.ts
pablitar: Yep. It is awaited here: https://github.com/decentraland/kernel/blob/c4437ba9c0423ff5fd57395cce8447b06ae9f667/packages/shared/dao/sagas.ts#L137

PR #10, opened 2021-08-19
refactor: change local storage to persistant storage

leamendoza97:
# Whats?
Change the name of atomicHelper module from 'localStorage' to 'persistentStorage'.
Also, it receives a custom storage in the initialization. Default persistantStorage is window.localStorage if it exists.
For preview.html it always mocks an arbitrary storage unless PLAYER query param be passed.

# Why?
localStorage is a specific implementation of HTML5 and it assumes you're running in a browser.
This change supports non-browser usages.


packages/atomicHelpers/persistentStorage.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  clear(): Promise<void> {
+    console.log("PersistentLocalStorage::clear")
+    return new Promise(resolve => {
+      this.storage.clear()
+      resolve()
+    })
+  }
```
menduz, line 21: For all async functions, there is no need to manually create `new Promise` if you add the `async` modifier to the function decl
```suggestion
  async clear(): Promise<void> {
    console.log("PersistentLocalStorage::clear")
    this.storage.clear()
  }
```

packages/shared/chat/sagas.ts
```diff
@@ -189,8 +189,9 @@ function initChatCommands() {
         response = TeleportController.goToMagic().message
       } else if (message.trim().toLowerCase() === 'random') {
         response = TeleportController.goToRandom().message
-      } else if (message.trim().toLowerCase() === 'next') {
-        response = TeleportController.goToNext().message
+      // TODO: this isn't a async or function generator...
+        // } else if (message.trim().toLowerCase() === 'next') {
+        // response = (await TeleportController.goToNext()).message
```
menduz, line 194: To wait for promises outside awaitable contexts, use .catch and .then
```ts
myAsyncFunction().catch(err => sharedLogger.error(err))
```

packages/shared/dao/sagas.ts
```diff
@@ -66,11 +66,11 @@ function* loadCatalystRealms() {
   if (!PREVIEW) {
     const network: ETHEREUM_NETWORK = yield select(getSelectedNetwork)
 
-    const cachedRealm: Realm | undefined = getFromLocalStorage(getLastRealmCacheKey(network))
+    const cachedRealm: Realm | undefined = yield getFromPersistentStorage(getLastRealmCacheKey(network))
```
menduz, line 69:
```suggestion
    const cachedRealm: Realm | undefined = yield call(getFromPersistentStorage, getLastRealmCacheKey(network))
```

To wait for async functions inside saga generators, please use `yield call(fn)` instead of `yield fn()`, it makes the code more testable. More info here: https://redux-saga.js.org/docs/advanced/Testing/, search for ".provide(" to view examples.

static/preview.html
```diff
@@ -551,6 +551,53 @@
...
+            return new Promise(resolve => {
+              resolve(key in storage ? storage[key] : null);
+            });
+          },
+          removeItem: function(key) {
+            return new Promise(resolve => {
+              delete storage[key];
+              update();
+              resolve()
+            });
```
menduz, line 585: This is another alternative implementation of the same code, without callbacks. No need to change it, just informative

```suggestion
            delete storage[key];
            update();
            return Promise.resolve()
```

packages/atomicHelpers/persistentStorage.ts
```diff
@@ -0,0 +1,97 @@
+import type { PersistentAsyncStorage } from "@dcl/kernel-interface"
+
+declare var window: any
+
+class PersistentLocalStorage implements PersistentAsyncStorage {
+  storage: any
+  constructor(storage: any){
```
menduz, line 7: Why does this constructo receives a "storage"? Should it have types?

menduz (changes requested):

packages/atomicHelpers/persistentStorage.ts
leamendoza97: That implementation of LocalPersistentStorage is a wrapper of localStorage. Maybe it should be in another file.
localStorage is a Storage type from DOM library.

LocalPersistentStorage keeps compatibility with kernel usages without storage option.

static/preview.html
leamendoza97: No need, but cleaner

packages/shared/dao/sagas.ts
leamendoza97: got it, ty

packages/shared/chat/sagas.ts
leamendoza97: This needs to change other functionality, it's called by a sync callback stored in a object. I need to be careful

leamendoza97: done

packages/shared/session/sagas.ts
```diff
@@ -161,12 +163,12 @@ function* authorize(requestManager: RequestManager) {
...
 
   if (isGuest) {
-    userData = getLastGuestSession()
+    userData = yield getLastGuestSession()
   } else {
     try {
       const address: string = yield getUserAccount(requestManager, false)
       if (address) {
-        userData = getStoredSession(address)
+        userData = yield getStoredSession(address)
```
menduz, line 171:
```suggestion
        userData = yield call(getStoredSession, address)
```

packages/shared/session/sagas.ts
```diff
@@ -161,12 +163,12 @@ function* authorize(requestManager: RequestManager) {
   const isGuest: boolean = yield select(getIsGuestLogin)
 
   if (isGuest) {
-    userData = getLastGuestSession()
+    userData = yield getLastGuestSession()
```
menduz, line 166:
```suggestion
    userData = yield call(getLastGuestSession)
```

packages/shared/world/TeleportController.ts
```diff
@@ -116,9 +116,9 @@ export class TeleportController {
...
 
-  public static goToNext(): { message: string; success: boolean } {
-    const current = getFromLocalStorage('launch-campaign-status') || 0
-    saveToLocalStorage('launch-campaign-status', current + 1)
+  public static async goToNext(): Promise<{ message: string; success: boolean }> {
+    const current = (await getFromPersistentStorage('launch-campaign-status')) || 0
+    await saveToPersistentStorage('launch-campaign-status', current + 1)
     const { x, y } = CAMPAIGN_PARCEL_SEQUENCE[current % CAMPAIGN_PARCEL_SEQUENCE.length]
     return TeleportController.goTo(x, y, `Teleporting you to the next scene... and more treasures!`)
   }
```
menduz, line 124: I think this is nolonger used. I would ask @nearnshaw and remove the code. That would simplify all the changes to `addChatCommand` enabling us to turning back those `async/await` changes

packages/entryPoints/index.ts
```diff
@@ -144,7 +149,7 @@ globalThis.DecentralandKernel = {
       async hasStoredSession(address: string, networkId: number) {
         if (!getStoredSession(address)) return { result: false }
```
menduz, line 151:
```suggestion
        if (!(await getStoredSession(address))) return { result: false }
```

PR #21, opened 2021-08-24
fix: biw correct name set in builder format

Suduck:
# What? 

Fixes decentraland/unity-renderer#950

This PR set the correct name for builder so when it publish the scene , it does it correctly


menduz (approved):

menduz:
@Suduck please verify in .zone in a couple of minutes, inform in #sdk-team if a full rollout is needed

PR #31, opened 2021-08-27
feat: export out of root folder 

leamendoza97:
# What? 
Change the url makers to add the pathname instead of only host.

# Why? 
This way allows serve the export from any path instead of only root.


static/export.html
```diff
@@ -566,7 +570,7 @@
           return `https://renderer-artifacts.decentraland.org/branch/${qs.get('renderer-branch')}`
         }
 
-        return new URL(`${location.origin}${location.pathname}unity-renderer`).toString()
+        return new URL(rootURL() + unity-renderer).toString()
```
menduz:
```suggestion
        return new URL(rootURL() + 'unity-renderer').toString()
```

packages/config/index.ts
```diff
@@ -141,13 +141,17 @@ export const FORCE_SEND_MESSAGE = location.search.includes('FORCE_SEND_MESSAGE')
...
 
-export const previewPinCatalyst = () => {
+export const rootURLPreviewMode = ({ endsWithSlash }: { endsWithSlash: boolean } = { endsWithSlash: false }) => {
   const url = `${location.origin}${location.pathname}`
-  return url.endsWith('/') ? url.substr(0, url.length-1) : url
+  if (url.endsWith('/')) {
+    return endsWithSlash ? url : url.replace(/$\//, '')
+  } else {
+    return !endsWithSlash ? url : `${url}/`
+  }
```
menduz: I rather create two functions with different names for this, the logic was very hard to follow without context

leamendoza97: Ok!, I think all this logic will need to move to some module in the future with a location abstraction.

nearnshaw (approved):

PR #41, opened 2021-09-06
feat: add video event handler and it's emitted to the owner scene

leamendoza97:
# What? 
Add video event handler and it's emitted to the owner scene.

# Why? 
https://github.com/decentraland/sdk/issues/11


menduz (approved):
:shipit:

PR #48, opened 2021-09-14
fix: Voice chat sometimes fails in Firefox

pablitar:
This PR fixes a corner case that would make Voice Chat completely fail on firefox sometimes.

This doesn't completely fix Voice Chat since there is an issue that I'm still investigating with web workers communication (and that seems to affect more things than just Voice Chat), but should improve the situation.

Also, we should get better connection results with less stun servers.


leamendoza97 (approved):
LGTM!

PR #53, opened 2021-09-20
fix: wait for renderer's DebugBridge ready before sending debug messages at initialization

pbosio:
Unity's receiver for debug messages is created at runtime, so we should wait until unity notify us that the receiver is up and ready


menduz (approved):
Cool, :shipit: 

I'm thinking about when we should use observables vs futures, let's try to figure out a way that makes sense to us for future PRs in pursuit of consistency in the code.

I think having futures scopes well the features but also couples the future to instances of classes, having global observables (or even sagas) may be better suited for most cases and for the overall decoupling of components

pbosio:
I got the same result in this PR https://github.com/decentraland/unity-renderer/pull/1226 without the need of any kernel modification. So I'm closing this :)

PR #54, opened 2021-09-21
chore: Added feature flag to allow unhealthy catalysts to be picked

pablitar:
Sometimes catalysts get out of sync with each other. This would make users stop picking these catalysts because it will be considered unhealthy.

If we are having a large amount of users and our catalysts start going unhealthy, we may have a cascading effect in which all catalysts get overloaded one by one until no catalyst is healthy and users cannot enter Decentraland. 

Even when a catalyst is unsynced, it may be able to serve users at least in a degraded form. This PR allows us to choose that behaviour by turning a feature flag on. Hopefully we don't need it, but just in case.


menduz:
Wow, it sounds like a really bad corner case, is this case even possible?

pablitar:
> Wow, it sounds like a really bad corner case, is this case even possible?
> 
> Did you consider first picking a healthy catalyst (if available) and then fallback to pick an unhealthy one? The behavior in this change also makes possible to connect to bootstrapping catalysts which is non desired at all.
> 
> Is there some RFC or more detailed information about why we should have this merged?

I'm not sure if this is possible. But we've seen a bunch of catalysts getting unhealthy all at once when the Metamask event happened.

We could filter bootstraping catalysts somehow, that'd probably be a good idea.

We could also prioritize healthy catalysts in the pick realm algorithm.

As the reasoning behind this, the idea is to have a kill switch in case something happens. Hopefully we don't have to use it. But if we have to...

PR #56, opened 2021-09-22
fix: test-scenes not loading

pbosio:
re-enable test-scenes for debugging.
using url param `ENABLE_TEST_SCENES`
also added url param `DISABLE_EMPTY_SCENES` to avoid loading empty-parcels (mostly for debugging porpoises)

### DIFF
--- a/packages/decentraland-loader/lifecycle/manager.ts
+++ b/packages/decentraland-loader/lifecycle/manager.ts
@@ -9,12 +9,14 @@ import { WebWorkerTransport } from 'decentraland-rpc/lib/common/transports/WebWo
 import { ensureMetaConfigurationInitialized } from 'shared/meta'
 import { getResourcesURL } from 'shared/location'
 
-import { parcelLimits, ENABLE_EMPTY_SCENES, LOS, getAssetBundlesBaseUrl } from 'config'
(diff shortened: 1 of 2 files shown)


packages/decentraland-loader/lifecycle/manager.ts
```diff
@@ -122,10 +124,11 @@ export async function initParcelSceneWorker() {
   const state = store.getState()
 
   const fullRootUrl = getResourcesURL('.')
+  const localServer = resolveUrl(`${location.protocol}//${location.hostname}:${8080}`, '/local-ipfs')
 
   server.notify('Lifecycle.initialize', {
-    contentServer: getFetchContentServer(state),
-    catalystServer: getCatalystServer(state),
+    contentServer: ENABLE_TEST_SCENES ? localServer : getFetchContentServer(state),
+    catalystServer: ENABLE_TEST_SCENES ? localServer : getCatalystServer(state),
```
menduz, line 131: Does it work without this? I don't remember what was `catalystServer` used for?

pbosio: It does not 😔
It won't load the test-scenes without that

menduz (approved):

menduz:
@Suduck @olavra

PR #62, opened 2021-09-28
refactor: automatic flag system

Suduck:
# What? 

This PR makes the flag system to be automatic so we don't need to add a PR to kernel each time a flag is added

# Why?

We wanted to eliminate the need of a PR in kernel each time a flag is added

### DIFF
--- a/packages/shared/meta/selectors.ts
+++ b/packages/shared/meta/selectors.ts
@@ -1,4 +1,4 @@
-import { BannedUsers, CommsConfig, FeatureFlags, MessageOfTheDayConfig, RootMetaState } from './types'
+import { BannedUsers, CommsConfig, MessageOfTheDayConfig, RootMetaState } from './types'
 import { Vector2Component } from 'atomicHelpers/landHelpers'
 import { AlgorithmChainConfig } from 'shared/dao/pick-realm-algorithm/types'
 import { DEFAULT_MAX_VISIBLE_PEERS } from '.'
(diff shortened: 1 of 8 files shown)


packages/unity-interface/kernelConfigForRenderer.ts
```diff
@@ -18,11 +15,7 @@ export function kernelConfigForRenderer(): KernelConfigForRenderer {
       nameValidCharacterRegex: nameValidCharacterRegex.toString().replace(/[/]/g, ''),
       nameValidRegex: nameValidRegex.toString().replace(/[/]/g, '')
     },
-    features: {
-      enableBuilderInWorld: false,
-      enableAvatarLODs: isFeatureEnabled(store.getState(), FeatureFlags.AVATAR_LODS, false),
-      enableExploreV2: false
-    },
+    features: {},
```
menduz, line 18: This is a breaking change, it will generate an imcompatibility conflict between versions of the renderer that assume `enableAvatarLODs: true` exist and now kernel will send `avatar_lods: true`

packages/shared/meta/selectors.ts
```diff
@@ -33,18 +33,44 @@ export const isMetaConfigurationInitiazed = (store: RootMetaState): boolean => s
...
+  store.meta.config.pickRealmAlgorithmConfig
 
 export const isMOTDInitialized = (store: RootMetaState): boolean =>
   store.meta.config.world ? store.meta.config.world?.messageOfTheDayInit || false : false
 export const getMessageOfTheDay = (store: RootMetaState): MessageOfTheDayConfig | null =>
   store.meta.config.world ? store.meta.config.world.messageOfTheDay || null : null
 
-export const isFeatureEnabled = (store: RootMetaState, featureName: FeatureFlags, ifNotSet: boolean): boolean => {
+export function getFeatureFlags(store: RootMetaState): Record<string, boolean> {
+  const featureFlags: Record<string, boolean> = {}
```
menduz, line 50: Please use `Record<string, FeatureFlagVariant>` instead of `Record<string, boolean>`

menduz (changes requested):
Please address the incompatibility issue in the code comment by adding a featuresV2 to the config map or similar approach, and replace `Record<string, boolean>` by `Record<string, FeatureFlagVariant>` which is broadly used by our feature flag systems. https://github.com/decentraland/feature-flags/blob/main/src/index.ts#L16-L39

packages/shared/meta/selectors.ts
Suduck: Right now I think that kernel is not using the feature flag library, should we implement the library and use it or we can copy the `FeatureFlagVariant ` type?

menduz: It can be copied, the datastructure in the getStore should be the same

menduz (approved):
If it works, merge it. Code looks good

PR #64, opened 2021-09-28
chore: Fixing double error handling so we can better report comms errors

pablitar:
We are reporting an error twice, and the second time it overrides the previous one, with something that's then reported incorrectly.

The error report that we are removing in this PR is then handled by this piece of code: https://github.com/decentraland/kernel/blob/5554c32bde2b54a0e966d30b98c3a1abc9282ce2/packages/shared/comms/index.ts#L1059

This will improve comms errors reporting, in conjunction with this other PR in website: https://github.com/decentraland/explorer-website/pull/76

image


menduz (approved):
I would approve this three times if it were possible, thanks :shipit:

PR #65, opened 2021-09-29
feat: add scene limitation warning for preview

pbosio:
* use `KernelConfigForRenderer` to enable `SceneDebugPanel`
* add to `KernelConfigForRenderer` option to enable scene limitation warning and to set `SceneDebugPanel` to be able to target a particular

### DIFF
--- a/packages/entryPoints/index.ts
+++ b/packages/entryPoints/index.ts
@@ -8,7 +8,7 @@ import { worldToGrid } from '../atomicHelpers/parcelScenePositions'
 import { DEBUG_WS_MESSAGES, ETHEREUM_NETWORK, HAS_INITIAL_POSITION_MARK, OPEN_AVATAR_EDITOR } from '../config/index'
 import 'unity-interface/trace'
 import { lastPlayerPosition, teleportObservable } from 'shared/world/positionThings'
-import { loadPreviewScene, startUnitySceneWorkers } from '../unity-interface/dcl'
+import { getPreviewSceneId, loadPreviewScene, startUnitySceneWorkers } from '../unity-interface/dcl'
 import { initializeUnity } from '../unity-interface/initializer'
 import { HUDElementID, RenderProfile } from 'shared/types'
 import { foregroundChangeObservable, isForeground } from 'shared/world/worldState'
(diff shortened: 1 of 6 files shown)


packages/entryPoints/index.ts
```diff
@@ -303,6 +303,19 @@ export async function startPreview() {
     wsScene = `${location.protocol === 'https:' ? 'wss' : 'ws'}://${document.location.host}/?scene`
   }
 
+  getPreviewSceneId().then((sceneData) => {
```
gonpombo8, line 306: can we use async/await here instead of promises.then ?
```ts
const sceneData = await getPreviewSceneId();
if (sceneData.sceneId) {
  const { unityInterface } = await ensureUnityInterface()
  unityInterface.SetKernel.....
}
```

packages/shared/types.ts
```diff
@@ -580,6 +580,13 @@ export type KernelConfigForRenderer = {
     enableBuilderInWorld: boolean
     enableAvatarLODs: boolean
   }
+  debugConfig:
+    | Partial<{
+        sceneDebugPanelEnabled: boolean
```
gonpombo8, line 585: we can remove the Partial with `|` and instead use the `?` operator.
```ts
debugConfig?: {
  sceneDebugPanelEnabled?: boolean
   sceneDebug...?: string | undefined
   etc
}
```

packages/unity-interface/dcl.ts
```diff
@@ -164,29 +165,34 @@ export async function startUnitySceneWorkers() {
...
 
-    // if (ws) {
-    //   transport = WebSocketTransport(new WebSocket(ws, ['dcl-scene']))
-    // }
+  if (sceneId) {
+    await reloadScene(sceneId)
   } else {
-    throw new Error('Could not load scene.json')
+    console.log(`Unable to load sceneId of ${sceneBase}`)
+    debugger
```
gonpombo8, line 188: do we still need this debugger here ?

packages/entryPoints/index.ts
pbosio: 👍 I will check if doing that change has an impact or not in the loading time of the preview scene. If there is any then I will change it as you suggest :)

packages/unity-interface/dcl.ts
pbosio: `sceneId` being `null` here means that there is an issue somewhere else, so it might become handy to have it there :)

packages/entryPoints/index.ts
```diff
@@ -303,6 +303,18 @@ export async function startPreview() {
...
+  getPreviewSceneId().then(async (sceneData) => {
+    if (sceneData.sceneId) {
+      const { unityInterface } = await ensureUnityInterface()
+      unityInterface.SetKernelConfiguration({
+        debugConfig: {
+          sceneDebugPanelTargetSceneId: sceneData.sceneId!,
+          sceneLimitsWarningSceneId: sceneData.sceneId!
+        }
+      })
+    }
```
menduz, line 315: I just saw where the bug of last week should have been resolved, lines https://github.com/decentraland/kernel/pull/65/files#diff-1200c56879b0f91125578151317e698e17f82d7d386495ca503416aa8fa45013R291-R293

 ```suggestion
      if(!queryString.has('position')){
           const position = pickWorldSpawnpoint(scene)
           i.Teleport(position)
           teleportObservable.notifyObservers(position.position)
      }
    }
```

Issue #63, opened 2021-09-28
Send "Version" in URL when requesting a profile from the Catalyst, if available


pablitar:
Profiles is one of the most frequent request that explorer makes while a person is using it.

Of the most frequent ones, is the one that we are not caching. This is because there is a particular common use case that'd be broken by caching:
1. I meet with my friends in world, to show them this cool wearable that I've got
2. I equip this wearable
3. My profile gets updated. I announce the update through comms.
4. My friends request the profile to the server. If we have caching, they could get an old version
5. They don't see I changed my wearable. Everybody is confused and unhappy :(

But there are ways to have caching and also support this use case: When a peer announces the profile modification through comms, it sends a "version", which is an incremental number. We can add that version to the query string of the request sent to the server, in order to change the URL. If the URL has changed, then it won't be cached by HTTP servers.

For instance: https://peer.decentraland.org/lambdas/profiles?id=0xA6CaB202588Bbd84bC3336C6B13Bb0542Bf0A6A7&version=15

This would be the first step. Once this has been implemented, we can change cache directives for the profiles endpoint without worrying about breaking this use case.

PR #67, opened 2021-10-01
chore: Requesting profiles using version

pablitar:
So we can use cache. This closes issue #63


menduz (approved):
Looking good, please merge it yourself when needed

PR #69, opened 2021-10-04
feat: add new message for requesting headers for builder

Suduck:
# What? 

This PR adds a new message for Unity to receive the requested headers for builder API.

It is generic enough to don't need more PRs if we want to add a new call

# Why? 

We need the headers of the builder API for a new call on Unity

### DIFF
--- a/packages/unity-interface/BrowserInterface.ts
+++ b/packages/unity-interface/BrowserInterface.ts
@@ -590,6 +590,17 @@ export class BrowserInterface {
     }
   }
 
+  public RequestHeaderForUrl(data: { method: string, url: string}) {
+    const identity = getCurrentIdentity(store.getState())
+    if (!identity) {
+      let emptyHeader: Record<string, string> = {}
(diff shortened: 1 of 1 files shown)


packages/unity-interface/BrowserInterface.ts
```diff
@@ -590,6 +590,17 @@ export class BrowserInterface {
     }
   }
 
+  public RequestHeaderForUrl(data: { method: string, url: string}) {
+    const identity = getCurrentIdentity(store.getState())
+    if (!identity) {
```
gonpombo8, line 595: u can avoid writing the same fn twice with a ternary, so if we need to change something in the future its easier 😄 

```ts
const headers: Record<string, string> = identity 
  ? BuilderServerAPIManager.authorize(identity, data.method, data.url)
  : {}
getUnityInstance().SendBuilderCatalogHeaders(headers)

```

Suduck: Changed!

PR #70, opened 2021-10-05
fix: change feature flag message order

Suduck:
# What? 

The feature flag should be sent before the asset bundle message since there is a debug system that is needed in order to receive the message to deactivate the assets bundles


menduz (dismissed):
Please add a comment to the code to prevent breaking this in the future

PR #74, opened 2021-10-06
add sfu voice store

gonpombo8:
# What? 
...

# Why? 
...


packages/shared/voice/sagas/signalConnection.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (voiceEnabled || true) { // TODO remove this
+      const { client, signal }: SignalConnection = yield call(createSignalConnection)
+      const socketChannel: EventChannel<VoiceActions> = yield call(createSocketChannel, { client, signal })
+
+      yield put(voiceInitialized(signal, client))
+      yield call(joinRoom)
+
+      while (true) {
+        try {
+          const payload: VoiceActions = yield take(socketChannel)
```
menduz, line 100: payload can be an error due to

```ts
    // If signal closes, then throw a new error.
    // The try/catch wrapper would do the reconnect workflow
    signal.onclose = (event) => {
      emit(new Error(event.type))
    }
```

is that intended?

packages/shared/voice/sagas/signalConnection.ts
```diff
@@ -0,0 +1,115 @@
...
+          yield put(payload)
+        } catch (error) {
+          // tslint:disable-next-line: no-console
+          console.log('Channel error:', error)
```
menduz, line 104:
```suggestion
          sharedLogger.log('Channel error:', error)
```

packages/shared/voice/sagas/signalConnection.ts
```diff
@@ -0,0 +1,115 @@
...
+          socketChannel.close()
+          throw error
+        }
+      }
+    }
+  } catch (error) {
+    // tslint:disable-next-line: no-console
+    console.log('Voice Saga:', error)
```
menduz, line 112:
```suggestion
    sharedLogger.log('Voice Saga:', error)
```

packages/shared/voice/sagas/signalConnection.ts
```diff
@@ -0,0 +1,115 @@
...
+  reconnectVoice
+} from '../actions'
+import { VoiceState } from '../types'
+import { joinRoom } from './joinRoom'
+
+type SignalConnection = Required<Pick<VoiceState, 'signal' | 'client'>>
+
+// TODO remove constants
+const ONE_MIMNUTE = 1000 * 60
+const URL = 'wss://test-sfu.decentraland.zone/ws'
```
menduz, line 23: please use the .services SFU, .zone doesn't work AFAIK

menduz (changes requested):
Is this ready for review? does it handle island changes?

packages/shared/voice/sagas/signalConnection.ts
gonpombo8: that emit new Error is handled by the `catch block`, so paylod can only be VoiceActions.

gonpombo8:
> Is this ready for review? does it handle island changes?
@menduz nope, its missing the entire VoiceContext part, but all the redux part its done. 

The island changes is handled here https://github.com/decentraland/kernel/pull/74/files#diff-f415fbae518cdf80eca665bfd3091096b819ad1d01bdc643c6cab48a3c6a32f1R21.

Also there is a bug or something i dont understand that leaving the current room its not working, so i had to do some trick (disconnect and connect the ws 😢). Im still looking for a way to resolve this.

PR #75, opened 2021-10-06
feat: Fetching ICE configuration from Catalyst

pablitar:
This PR makes a request to the `peer-parameters` endpoint, which will respond with parameters corresponding to the current user.

Currently, it only responds ICE servers. We can use this to keep working on configuring turn server without needing to deploy new code.

### DIFF
--- a/packages/atomicHelpers/signedFetch.ts
+++ b/packages/atomicHelpers/signedFetch.ts
@@ -1,0 +1,52 @@
+import { AuthChain, Authenticator, AuthIdentity } from "dcl-crypto"
+import { flatFetch, FlatFetchInit } from "./flatFetch"
+
+const AUTH_CHAIN_HEADER_PREFIX = 'x-identity-auth-chain-'
+const AUTH_TIMESTAMP_HEADER = 'x-identity-timestamp'
+const AUTH_METADATA_HEADER = 'x-identity-metadata'
+
+function getAuthHeaders(
+  method: string,
+  path: string,
+  metadata: Record<string, any>,
+  chainProvider: (payload: string) => AuthChain
+) {
+  const headers: Record<string, string> = {}
+  const timestamp = Date.now()
+  const metadataJSON = JSON.stringify(metadata)
+  const payloadParts = [method.toLowerCase(), path.toLowerCase(), timestamp.toString(), metadataJSON]
+  const payloadToSign = payloadParts.join(':').toLowerCase()
+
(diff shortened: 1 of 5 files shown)


packages/shared/comms/index.ts
```diff
@@ -890,11 +891,33 @@ function subscribeToRealmChange(store: Store<RootState>) {
   )
 }
 
+async function getPeerParameters(): Promise<PeerParameters> {
+  const commsServer = getCommsServer(store.getState())
```
menduz, line 895: Can we move the endpoint to a lambda? To decouple the configuration of comms from the actual implementation

packages/config/index.ts
```diff
@@ -171,10 +171,10 @@ export namespace commConfigurations {
   export const autoChangeRealmInterval =
     typeof qs.AUTO_CHANGE_INTERVAL === 'string' ? parseInt(qs.AUTO_CHANGE_INTERVAL, 10) * 1000 : 40000
 
-  export const iceServers = [
+  export const defaultIceServers = [
     { urls: 'stun:stun.l.google.com:19302' },
     {
-      urls: 'turn:stun.decentraland.org:3478',
+      urls: 'turn:coturn-raw.decentraland.services:3478',
```
menduz, line 177: I'd remove turn from the defaults. Keeping only stun. The .services domain is ephemeral

packages/atomicHelpers/signedFetch.ts
```diff
@@ -0,0 +1,52 @@
...
+    headers[`${AUTH_CHAIN_HEADER_PREFIX}${index}`] = JSON.stringify(link)
+  })
+
+  headers[AUTH_TIMESTAMP_HEADER] = timestamp.toString()
+  headers[AUTH_METADATA_HEADER] = metadataJSON
+
+  return headers
+}
+
+export function signedFetch(url: string, identity: AuthIdentity, init?: FlatFetchInit, additionalMetadata: Record<string, any> = {}) {
```
menduz, line 32: Cool

menduz (changes requested):

menduz (approved):

packages/shared/comms/index.ts
pablitar: I see. Still, I think we should add those endpoints as part of solving the particular problems they will solve. Mixing problems can lead us to try and solve something that we don't yet have defined.

Thanks for approving!

PR #76, opened 2021-10-08
chore: reorg comms, prepare for new versions

menduz:
Pending: 
- add https://github.com/decentraland/kernel/pull/112/files
- add https://github.com/decentraland/kernel/pull/112/files
-


packages/shared/comms/context.ts
```diff
@@ -0,0 +1,445 @@
...
+  }
+}
+
+let currentParcelTopics = ''
+let previousTopics = ''
+
+let lastNetworkUpdatePosition = new Date().getTime()
+let lastPositionSent: Position | undefined
+
+function onPositionUpdate(context: CommsContext, p: Position) {
```
pablitar, line 361: Should this be in context.ts? Or should this be in handlers.ts?

packages/shared/comms/handlers.ts
```diff
@@ -0,0 +1,253 @@
+import { genericAvatarSnapshots, COMMS_PROFILE_TIMEOUT } from 'config'
```
pablitar, line 1: I'm assuming this code was moved without changes. If there is a change, please let me know!

packages/shared/comms/context.ts
```diff
@@ -0,0 +1,445 @@
+import { commConfigurations, parcelLimits } from 'config'
```
pablitar, line 1: I'm assuming most of this code was moved without changes, other than converting some global functions into methods of comms context.

If there's something more significant than that that I should review, please let me know!

packages/shared/dao/types.ts
```diff
@@ -33,19 +32,14 @@ type BaseCandidate = {
...
-
 export type IslandsBasedCandidate = {
   type: 'islands-based'
   usersCount: number
   usersParcels?: Parcel[]
   maxUsers?: number
 } & BaseCandidate
 
-export type Candidate = LayerBasedCandidate | IslandsBasedCandidate
+export type Candidate = IslandsBasedCandidate
```
pablitar, line 42: Maybe we can remove the type `IslandBasedCandidate`?

packages/shared/comms/index.ts
```diff
@@ -896,45 +106,60 @@ async function getPeerParameters(): Promise<PeerParameters> {
...
     }
 
+    const user = await getStoredSession(identity?.address)
+
+    if (!user) {
+      return
+    }
+    
     const userInfo = {
       userId,
```
pablitar, line 146:
```suggestion
      userId: identity.address,
```

packages/shared/dao/types.ts
menduz: 👍

PR #77, opened 2021-10-08
Message with unity renderer to update profile description

lorux0:
# What?
Message received from unity renderer to update the **profile description**.

# Why?
It is required in unity renderer to be able to update profile description from the profile menu.


packages/unity-interface/BrowserInterface.ts
```diff
@@ -309,6 +309,10 @@ export class BrowserInterface {
   public SaveUserUnverifiedName(changes: { newUnverifiedName: string }) {
     store.dispatch(saveProfileRequest({ unclaimedName: changes.newUnverifiedName }))
   }
+  
+  public SendSaveUserDescription(changes: { description: string }) {
```
pbosio, line 313:
```suggestion
  public SaveUserDescription(changes: { description: string }) {
```

PR #83, opened 2021-10-18
feat: add method to get portable experiences loaded in PortableExperi…

leanmendoza:
…ence API

# What? 
...

# Why? 
...

### DIFF
--- a/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
+++ b/packages/shared/apis/SceneStateStorageController/BuilderServerAPIManager.ts
@@ -113,7 +113,7 @@ export class BuilderServerAPIManager {
       const response = await fetch(urlToFecth, params)
       const data = await response.json()
 
-      const manifest: BuilderManifest = data.data
+      const manifest: BuilderManifest = data
 
       // If this manifest contains assets, we add them so we don't need to fetch them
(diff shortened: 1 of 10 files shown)


packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -133,6 +139,17 @@ export async function getLoadablePortableExperience(data: {
...
 }
 
+export async function getLoadedPortableExperiences() {
+  let portableExperiences:string[] = []
+  await loadedSceneWorkers.forEach(async (worker, key) => {
+    const parcelIdentity = await worker.getAPIInstance(ParcelIdentity)
+    if (parcelIdentity.isPortableExperience){
+      portableExperiences.push(key)
+    }
+  })
```
menduz, line 149:
```suggestion
  for (const [worker, key] of loadedSceneWorkers) {
    const parcelIdentity = await worker.getAPIInstance(ParcelIdentity)
    if (parcelIdentity.isPortableExperience){
      portableExperiences.push(key)
    }
  }
```

menduz (changes requested):
await map.forEach does not work as expected. It should be `await Promise.all(...`

menduz (approved):

PR #86, opened 2021-10-19
feat: add emit profile changed event

leanmendoza:
move allSceneEvent to parcelSceneManager

# What? 
...

# Why? 
...

### DIFF
--- a/packages/shared/profiles/sagas.ts
+++ b/packages/shared/profiles/sagas.ts
@@ -65,6 +65,7 @@ import { ParcelsWithAccess } from 'decentraland-ecs'
 import { getUnityInstance } from 'unity-interface/IUnityInterface'
 import { store } from 'shared/store/isolatedStore'
 import { createFakeName } from './utils/fakeName'
+import { allScenesEvent } from 'shared/world/parcelSceneManager'
 
 const toBuffer = require('blob-to-buffer')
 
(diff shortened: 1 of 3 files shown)


packages/shared/profiles/sagas.ts
```diff
@@ -95,7 +96,7 @@ export function* profileSaga(): any {
   yield takeEvery(USER_AUTHENTIFIED, initialProfileLoad)
 
   yield takeLatestByUserId(PROFILE_REQUEST, handleFetchProfile)
-  yield takeLatestByUserId(PROFILE_SUCCESS, submitProfileToRenderer)
+  yield takeLatestByUserId(PROFILE_SUCCESS, submitProfileToRendererAndScenes)
```
menduz, line 99: I think you should hook to `SAVE_PROFILE_SUCCESS` (after self profile changed) instead of `PROFILE_SUCCESS` (after all avatars either load or change)

packages/shared/world/parcelSceneManager.ts
```diff
@@ -216,3 +217,9 @@ export async function enableParcelSceneLoading(options: EnableParcelSceneLoading
     ret.notify('User.setPosition', { position: obj.newParcel, teleported: false })
   })
 }
+
+export function allScenesEvent(data: { eventType: string; payload: any }) {
+  for (const [_key, scene] of loadedSceneWorkers) {
+    scene.emit(data.eventType as IEventNames, data.payload)
+  }
+}
```
menduz, line 225: cool

packages/shared/profiles/sagas.ts
leanmendoza: ok, I thought we wanted to know when every profile changes and even load. For example if I'm running the scene A and this scene wants to match me with users that don't have any portable experience. It will only know my changes with SAVE_PROFILE_SUCCESS

PR #89, opened 2021-10-20
fix: removes duplicate names from explorer builder translation

Suduck:
# What? 

This PR makes the entities from the builder have a unique name. 

# Why?

The builder uses the names from the entities to publish the scene if several items have the same name, they won't be published.

In order to increase the compatibility between builder and explorer builder this PR address this problem


packages/shared/apis/SceneStateStorageController/utils.ts
```diff
@@ -16,10 +16,23 @@ const HUMAN_READABLE_TO_ID: Map<string, number> = new Map([
   ['Script', CLASS_ID.SMART_ITEM]
 ])
 
+export function getUniqueNameForGLTF(currentNames: String[], gltfName: string, amountOfTimesAppear: number): string {
+  let nameToReturn: string = gltfName
+
+  if (amountOfTimesAppear > 1) nameToReturn = nameToReturn + amountOfTimesAppear
+
+  currentNames.forEach((name) => {
```
menduz, line 24: Use for of instead of foreach when possible

menduz (dismissed):

BrianAmadori (dismissed):

PR #98, opened 2021-11-01
feat: [Explore V2] Signed requests to Events API (Kernel side)

sandrade-dcl:
# What? 
Make signed requests (through `signedFetch`) to the Events API.

# Why? 
Once the new version of the Events API is available, we will need to sign the requests in order to be able to obtain custom user fields like `attending`  that will inform us if out user is subscribed to an event or not.

### DIFF
--- a/packages/unity-interface/BrowserInterface.ts
+++ b/packages/unity-interface/BrowserInterface.ts
@@ -674,8 +674,8 @@ export class BrowserInterface {
   }
 
   public async RequestDCLEvents(urlPayload: { value: string }) {
-    var eventListPayload = null
-    var isOk = false
+    let eventListPayload = null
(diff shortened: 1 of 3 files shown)


menduz (changes requested):
Please do expose signedFetch to unity instead. That can be leveraged for other future integrations

pbosio:
@sandrade-dcl @menduz is this still active or can be closed?

sandrade-dcl:
> @sandrade-dcl @menduz is this still active or can be closed?

This PR was in progress but freezed because we're waiting for the new DCL Events API is implemented. Anyway, I think this implementation is very outdated and we can close the PR at the moment. When we resume this development we'll start it from scratch (there was little developed).

PR #102, opened 2021-11-04
feat: extend getUserData to fetch data from other users

pbosio:


leanmendoza (approved):

menduz (changes requested):
Hi Pato, I have one comment and one question

1. The JSON-RPC protocol does not allow optional params, and even though the typings are correct, it changes the signature breaking the motto "if the function does or return something different, it should be called differently". To prevent future incompatibilities, this time should add a new function to perform the query for other players.
2. What is the "avatar" field supposed to be and why was it added to the returned value?

menduz (approved):

PR #105, opened 2021-11-12
feat: avatar sync in preview mode (comms/v1)

leanmendoza:
# What? 
Add profile request and profile response messages through comms. Also, fix the correspondence between ProfileType messages and profile module.

# Why? 
This fix https://github.com/decentraland/sdk/issues/43


packages/shared/comms/v1/proto/comms.proto
```diff
@@ -39,12 +41,25 @@ message ProfileData {
     double time = 2;
     string profile_version = 3;
     enum ProfileType {
-        DEPLOYED = 0;
-        LOCAL = 1;
+        LOCAL = 0;
+        DEPLOYED = 1;
```
menduz, line 45: How does this affect to the current comms over WebRTC?

menduz (approved):

packages/shared/comms/v1/proto/comms.proto
leanmendoza: Mmm, while I was debugging this, I found an incoherence with https://github.com/decentraland/kernel/blob/main/packages/shared/profiles/types.ts#L59
I didn't modify the v2 comms but, the incoherence exits there too.
Maybe it worths to take a look.

leanmendoza: But, this change shouldn't modify the behaviour of comms over WebRTC

menduz: 👌

PR #106, opened 2021-11-15
feat: add catalog source for wearable in preview mode

leanmendoza:
# What? 
...

# Why? 
...


packages/shared/catalogs/sagas.ts
```diff
@@ -155,6 +156,11 @@ function* fetchWearablesFromCatalyst(filters: WearablesRequestFilters) {
     }
   }
 
+  if (PREVIEW && COLLECTIONS_ALLOWED) {
```
menduz, line 159: Is the `COLLECTIONS_ALLOWED` also required?

menduz (approved):

PR #108, opened 2021-11-18
feat: add event for peer's avatar connected or disconnected

pbosio:
+add `getConnectedPlayers` to fetch from a scene all connected and visible avatars
+add `avatarTracker` that send events to the scenes when an avatar spawns or is removed


menduz (dismissed):

menduz (approved):

packages/shared/social/avatarTracker.ts
```diff
@@ -0,0 +1,48 @@
+import { AvatarMessageType } from 'shared/comms/interface/types'
+import { avatarMessageObservable } from 'shared/comms/peers'
+import { allScenesEvent } from 'shared/world/parcelSceneManager'
+
+const avatarConnected = 'playerConnected'
+const avatarDisconnected = 'playerDisconnected'
+
+const avatarMap: Record<string, { userId: string; visible: boolean }> = {}
```
gonpombo8, line 8: Can we use a Set of visible avatars instead of saving if the user is visible or not.
And do the mapping of uuid -> userId with the getUser fn that comms exports.
Im thinking something more readable like this.
```ts
const avatarSet: Set<string> = new Set()
export function getVisibleAvatarsUserId(): string[] {
  return Array.from(avatarSet).map(uuid => getUser(uuid)?.userId)
}

avatarMessageObservable.add((evt) => {
  if (evt.type === AvatarMessageType.USER_VISIBLE) {
    const visible = avatarSet.has(evt.uuid)

    if (visible !== evt.visible) {
      const userId = getUser(evt.uuid)?.userId

      if (!userId) return

      allScenesEvent({
        eventType: evt.visible ? avatarConnected : avatarDisconnected,
        payload: { userId }
      })

      if (!evt.visible) {
        avatarSet.delete(evt.uuid)
        return
      }

      avatarSet.add(evt.uuid)
    }
  }
})
```

What do u think? If you dont agree or find something strange you can avoid the comment.
Just some personal `tocs` 😆

packages/shared/social/avatarTracker.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (evt.type === AvatarMessageType.USER_VISIBLE) {
+    if (avatarMap[evt.uuid]) {
+      const isVisible = avatarMap[evt.uuid].visible
+      const userId = avatarMap[evt.uuid].userId
+
+      if (!userId) {
+        return
+      }
+
+      if (isVisible != evt.visible) {
```
gonpombo8, line 26: `!==` use strict comparison

gonpombo8 (changes requested):

packages/shared/social/avatarTracker.ts
pbosio: the main problems I see there are:

1.
```
export function getVisibleAvatarsUserId(): string[] {
  return Array.from(avatarSet).map(uuid => getUser(uuid)?.userId)
}
```
if `getUser(uuid)` is undefined or `userId` is undefined, wouldn't it be filling the string array with undefined elements?

2.
```const userId = getUser(evt.uuid)?.userId```
I would also have check for `evt.type === AvatarMessageType.USER_REMOVED` but peer's uuid is removed  from the map before triggering the event (https://github.com/decentraland/kernel/blob/main/packages/shared/comms/peers.ts#L48-L53)

So I'm not entirely sure if I can remove the uuid <> userId mapping in this file... but I'll se what I can do to make it more readable

Comment (gonpombo8) [reply packages/shared/social/avatarTracker.ts:8]: USER_REMOVED doesn't trigger USER_INVISIBLE with false ? 

If thats the case then we should do the uuid <> userId mapping 😢. But I think that the USER_DATA event can be removed and use the getPeer fn instead to do this. 
```ts
const avatarMap = new Map<string, string>()
avatarMap.add(uuid, addressId)
```

gonpombo8 (approved):
❤️

PR #110, opened 2021-11-18
fix: deprecate realm's layer from EnvironmentAPI

pbosio:
*deprecated `layer` from `EnvironmentRealm`:
*`layer` is now set with the `island` id
+added `room` field that is set with the `island` id


pbosio:
should we address this issue https://github.com/decentraland/sdk/issues/132 before merging this PR?

gonpombo8 (approved):

PR #111, opened 2021-11-19
fix: WITH_COLLECTIONS query param base wearables fetch

pravusjif:
### WHY
When opening the client with `WITH_COLLECTIONS` query param, if the user doesn't have a profile yet, the wearables request will wait forever.

When that happens the UX is as follows:
1. When entering as an incognito guest the backpack is the first thing you see to create the character
2. The backpack appears totally empty (no base wearables either)
3. After several wearables requests retries from the client (45 secs) the explorer crashes

Issue solved by this PR: https://github.com/decentraland/unity-renderer/issues/1547

### WHAT
Removed unneeded yield that in some occasions breaks the base wearables fetch and thus the client explodes

### TEST
**Production (bugged)**
Enter this URL **in incognito + guest mode**: https://play.decentraland.zone/?&NETWORK=ropsten&WITH_COLLECTIONS=3062136a-065d-4d94-b28c-f57d6ef04860&island=I3jg&position=0%2C-1&realm=thor&DEBUG_MODE

**This PR (fixed)**
Enter this URL **in incognito + guest mode**: https://play.decentraland.zone/?kernel-branch=fix%2FWithCollectionsParamBaseWearablesFetch&NETWORK=ropsten&WITH_COLLECTIONS=3062136a-065d-4d94-b28c-f57d6ef04860&island=I3jg&position=0%2C-1&realm=thor&DEBUG_MODE

### DIFF
--- a/packages/shared/catalogs/sagas.ts
+++ b/packages/shared/catalogs/sagas.ts
@@ -36,7 +36,6 @@ import { getCatalystServer, getFetchContentServer, getSelectedNetwork } from 'sh
 import { BuilderServerAPIManager } from 'shared/apis/SceneStateStorageController/BuilderServerAPIManager'
 import { getCurrentIdentity } from 'shared/session/selectors'
 import { getUnityInstance } from 'unity-interface/IUnityInterface'
-import { onLoginCompleted } from 'shared/session/sagas'
 import { ExplorerIdentity } from 'shared/session/types'
 
 export const BASE_AVATARS_COLLECTION_ID = 'urn:decentraland:off-chain:base-avatars'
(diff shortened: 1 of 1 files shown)


pbosio:
additional information about this issue:
when opening the client with `WITH_COLLECTIONS` query param, if user doesn't have a profile yet, the wearables request will wait forever on this instruction: `yield onLoginCompleted()` because `identity` does not exist in the store until the "sign up" flow is completed.
this PR fix the client crashing, after a timeout, due to the lack of wearables

gonpombo8 (approved):

PR #112, opened 2021-11-22
feat: add realm changed event for sdk

pbosio:
send `onRealmChanged` to all scenes when realm and/or island is set


gonpombo8 (approved):

gonpombo8:
Im thinking that maybe we could have a `sagas/scene-events.ts` and connect here all the events that we are currently sending to the scene.
So this way we dont have this logic separated, and every time we need to add a new event, we know that it should be there.

PR #113, opened 2021-11-23
add posix and update ecs version to 6.6.9

gonpombo8:
# What? 
...

# Why? 
...


nearnshaw:
I think this next version should be called 6.7.0, not 6.6.9
We're adding quite a few features and changes, it's worth the second digit :)

menduz:
@nearnshaw The number here has nothing to do with the actual version. That is handled in js-sdk-toolchain, kernel is a different component and should be upgraded to the most recent version

PR #117, opened 2021-11-26
feat: add enter and leave scene events for remote user

pbosio:
+report player enter/leave scene
+add `getPlayersInScene` to fetch players currently inside scene

### DIFF
--- a/packages/shared/apis/Players.ts
+++ b/packages/shared/apis/Players.ts
@@ -1,5 +1,6 @@
 import { registerAPI, exposeMethod } from 'decentraland-rpc/lib/host'
 import { ExposableAPI } from './ExposableAPI'
+import { ParcelIdentity } from './ParcelIdentity'
 
 import { store } from 'shared/store/isolatedStore'
 
@@ -8,7 +9,7 @@ import { hasConnectedWeb3 as hasConnectedWeb3Selector } from 'shared/profiles/se
 import { getProfileIfExist } from 'shared/profiles/ProfileAsPromise'
 import { calculateDisplayName } from 'shared/profiles/transformations/processServerProfile'
 
-import { getVisibleAvatarsUserId } from 'shared/social/avatarTracker'
+import { getVisibleAvatarsUserId, getInSceneAvatarsUserId } from 'shared/social/avatarTracker'
 
 export interface IPlayers {
   /**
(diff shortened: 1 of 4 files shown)


packages/shared/social/avatarTracker.ts
```diff
@@ -1,17 +1,32 @@
...
 
 export function getVisibleAvatarsUserId(): string[] {
   return Array.from(visibleAvatars.values())
 }
 
+export function getInSceneAvatarsUserId(sceneId: string): string[] {
+  return Array.from(rendererAvatars.entries())
+    .filter(([userId, avatarData]) => avatarData.sceneId === sceneId)
+    .map(([userId, avatarData]) => userId)
+}
```
menduz, line 18: The functional approach is heavy for the garbage collector and should be avoided when performance is critical.. 

```suggestion
export function* getInSceneAvatarsUserId(sceneId: string): Iterable<string> {
  for (const [userId, avatarData] of rendererAvatars) {
    if (avatarData.sceneId === sceneId) yield userId
  }
}
```

Then can be used with a `for` (optimal) or `Array.from(getInSceneAvatarsUserId())`

packages/shared/apis/Players.ts
```diff
@@ -47,4 +52,23 @@ export class Players extends ExposableAPI implements IPlayers {
...
+      return { userId }
+    })
+
+    // check also for current user, since it will appear in `getInSceneAvatarsUserId` result
+    if (isWorldPositionInsideParcels(parcelIdentity.land.sceneJsonData.scene.parcels, lastPlayerPosition)) {
+      const currentUserId = getCurrentUserId(store.getState())
+      if (currentUserId) {
+        result.push({ userId: currentUserId })
+      }
+    }
```
menduz, line 70: How can we be sure to not be pusing the userId twice? where is the logic that prevents `result` from containing our userId?

menduz (changes requested):

packages/shared/apis/Players.ts
pbosio: Renderer won't report current user's avatar since it is only reporting `AvatarShape` and the avatar for current user is not making use of an `AvatarShape`. I could add a check if `userId` is already included if you want

menduz: Ok sounds good. I think a that would be a good defensive measure 👍

menduz (approved):

PR #120, opened 2021-12-02
chore: add world config to kernel config

BrianAmadori:
# What? 
This PR makes the `world` object of the `meta` configuration file pass through directly to renderer, without having to declare new members explicitly on the `WorldConfig` type.

# Why? 
We want to be able to add new configuration to renderer without having to open PRs anymore. This eventually may be deprecated in favor of feature variants. This is currently blocking the procedural skybox project.

### DIFF
--- a/packages/shared/meta/types.ts
+++ b/packages/shared/meta/types.ts
@@ -17,7 +17,7 @@ export type MetaConfiguration = {
   pickRealmAlgorithmConfig?: AlgorithmChainConfig
   bannedUsers: BannedUsers
   synapseUrl: string
-  world: WorldConfig
+  world: any
   comms: CommsConfig
   minCatalystVersion?: string
   featureFlags?: Record<string, boolean>
--- a/packages/shared/meta/types.ts
+++ b/packages/shared/meta/types.ts
@@ -17,7 +17,7 @@ export type MetaConfiguration = {
   pickRealmAlgorithmConfig?: AlgorithmChainConfig
   bannedUsers: BannedUsers
   synapseUrl: string
-  world: any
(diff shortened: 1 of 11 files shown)


packages/shared/meta/types.ts
```diff
@@ -17,7 +17,7 @@ export type MetaConfiguration = {
   pickRealmAlgorithmConfig?: AlgorithmChainConfig
   bannedUsers: BannedUsers
   synapseUrl: string
-  world: WorldConfig
+  world: any
```
menduz, line 20: This is the only questionable line of the PR. Could you add a rationale?

menduz (changes requested):

packages/shared/meta/types.ts
BrianAmadori: In the json, the world object can have more fields, but those fields are only going to be needed renderer side, so to avoid having to create a new PR on Kernel every time a new field is added to the json file we treat this as `any` and pass it over directly to Unity, where it will be statically typed as needed. If we don't do this, the extra fields are lost because of the explicit typing.

To avoid treating this as `any` everywhere, I added a `getWorldConfig()` selector so we can still have static typing through the `WorldConfig` type, with only for the subset fields that are used by Kernel.

BrianAmadori: Ok, I ran this through fiddle and I think the `any` type is not needed, but I'll have to double check /shrug

BrianAmadori: Everything works fine with the static typing, done.

menduz (approved):

PR #121, opened 2021-12-08
add allSceneEvents typings

gonpombo8:
# What? 

Add typings to the events that we send to the scene.
...

# Why? 

To avoid typos or sending invalid payloads.

image

...


pbosio (approved):
👏

PR #122, opened 2021-12-09
feat: isolated scenes

Suduck:
# What? 

This PR implements the isolated mode for kernel.

# Why? 

It is explained in this ADR https://github.com/decentraland/adr/pull/31


packages/shared/world/types.ts
```diff
@@ -0,0 +1,21 @@
+import { ILand } from 'shared/types'
+
+export enum IsolatedMode {
+  BUILDER = 0 // Payload to use: BuilderIsolatedModeOptions
+}
+
+export type IsolatedModeOptions = {
```
gonpombo8, line 7: To avoid doing `as SomeType` when we want to cast the payload.
```suggestion
export type IsolatedModeOptions<T = any> = {
  mode: IsolatedMode
  payload: T
}
```

packages/shared/world/parcelSceneManager.ts
```diff
@@ -122,99 +145,260 @@ function reportPendingScenes() {
...
-    const sceneId = opts.sceneId
-    const parcelSceneToStart = await ret.getParcelData(sceneId)
+  // create the worker if don't exis
+  if (!getSceneWorkerBySceneID(sceneId)) {
+    //If we are running in isolated mode and it is builder mode, we create a stateless worker instead of a normal worker
+    if (
+      parcelSceneLoadingState.runningIsolatedMode &&
+      parcelSceneLoadingState.isolatedModeOptions?.mode === IsolatedMode.BUILDER
+    ) {
+      let sceneId: string = ''
```
gonpombo8, line 291: same here, this could be a const with a ternary so we avoid reading an if with a mutation after

packages/shared/world/parcelSceneManager.ts
```diff
@@ -122,99 +145,260 @@ function reportPendingScenes() {
...
-      options.onLoadParcelScenes([parcelSceneToStart])
+  timer = setTimeout(() => {
+    const worker = getSceneWorkerBySceneID(sceneId)
+    if (worker && !worker.hasSceneStarted()) {
+      sceneLifeCycleObservable.remove(observer)
+      worker.ready |= SceneWorkerReadyState.LOADING_FAILED
+      lifecycleManager.notify('Scene.status', { sceneId, status: 'failed' })
+      globalSignalSceneFail(sceneId)
     }
+  }, 90000)
```
gonpombo8, line 341: avoid magic numbers
```ts
const WORKER_TIMEOUT = 90000
```

gonpombo8 (commented):
could we add some tests for this since now they are working as expected ?
Maybe to the start & stop to check that desiredScenes set its being updated as expected.

packages/shared/world/StatefulWorker.ts
```diff
@@ -4,19 +4,21 @@ import { SceneWorker } from './SceneWorker'
...
+import { ParcelIdentity } from 'shared/apis/ParcelIdentity'
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const gamekitWorkerRaw = require('raw-loader!../../../static/systems/stateful.scene.system.js')
 const gamekitWorkerBLOB = new Blob([gamekitWorkerRaw])
 const gamekitWorkerUrl = URL.createObjectURL(gamekitWorkerBLOB)
 
 export class StatefulWorker extends SceneWorker {
-  constructor(parcelScene: ParcelSceneAPI) {
+  constructor(parcelScene: ParcelSceneAPI, isEmptyStatefull: boolean) {
```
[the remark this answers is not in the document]
Suduck: done

packages/shared/world/parcelSceneManager.ts
```diff
@@ -122,99 +145,260 @@ function reportPendingScenes() {
   store.dispatch(informPendingScenes(pendingScenes.size, countableScenes))
 }
 
-export async function enableParcelSceneLoading(options: EnableParcelSceneLoadingOptions) {
-  const ret = await initParcelSceneWorker()
+const parcelSceneLoadingState = {
```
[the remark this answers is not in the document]
Suduck: Added the type for the parcelSceneLoadingState, for the payload, it can be different for each mode, right now there is only the builder mode but in the future, it can have different values, we still want to type it?

gonpombo8 (approved):

PR #124, opened 2021-12-10
feat: add permissions module

leanmendoza:
# What? 
...

# Why? 
...


packages/shared/apis/EthereumController.ts
```diff
@@ -49,26 +50,38 @@ export interface IEthereumController {
...
     return sendAsync(message)
   }
 
   @exposeMethod
   async getUserAccount(): Promise<string | undefined> {
+    await this.ensureHasPermissions()
     return getUserAccount(requestManager)
   }
+
+  private async ensureHasPermissions() {
```
gonpombo8, line 81: could this fn be at the `ExposableAPI` ? So we dont need to re-write it on everey api we create.
Also it could receive an array of permissions.
```ts
private async ensureHasPermissions(permisions: keyofPermissions[])
```

packages/shared/apis/Permissions.ts
```diff
@@ -0,0 +1,39 @@
...
+   */
+  @exposeMethod
+  async hasPermission(test: PermissionItem): Promise<boolean> {
+
+    // Backward compatibility with parcel scene with 'requiredPermissions' in the scene.json
+    //  Only the two permissions that start with ALLOW_TO_... can be conceed without user
+    //  interaction
+    const json = this.parcelIdentity.land.sceneJsonData
+    const list = json.requiredPermissions || []
+    if (list.indexOf(test) !== -1 &&
```
gonpombo8, line 30: `list.includes(test)` just a tock with indexOf. I think cause its more readable

gonpombo8 (changes requested):

packages/shared/apis/EthereumController.ts
leanmendoza: Maybe we can create another ExposableAPI like RestrictedExposableAPI to give this default methods?

packages/shared/apis/Permissions.ts
```diff
@@ -26,14 +26,21 @@ export class Permissions extends ExposableAPI {
...
 
-    return this.permissionGranted.indexOf(test) !== -1
+    return this.permissionGranted.includes(test)
+  }
+  /**
+   * Returns if it has many permissions
+   */
+  @exposeMethod
+  async hasManyPermissions(test: PermissionItem[]): Promise<boolean> {
+    return test.filter(item => this.hasPermission(item)).length == test.length
```
gonpombo8, line 43: this could be replaced with a `every`. And also rename the param to what it is 😛 
```ts
return permissionItems.every(item => this.hasPermission(item))
```

gonpombo8 (changes requested):

packages/shared/apis/Permissions.ts
leanmendoza: I replaced it with explicit for, it's more readable (because the promises, await, etc)

gonpombo8 (approved):

packages/shared/apis/RestrictedActions.ts
```diff
@@ -62,10 +59,7 @@ export class RestrictedActions extends ExposableAPI {
   @exposeMethod
   async triggerEmote(emote: Emote): Promise<void> {
     // checks permissions
-    if (!this.hasPermission(PermissionItem.ALLOW_TO_TRIGGER_AVATAR_EMOTE)) {
-      defaultLogger.error(`Permission "${PermissionItem.ALLOW_TO_TRIGGER_AVATAR_EMOTE}" is required`)
-      return
-    }
+    await this.ensureHasPermissions([PermissionItem.ALLOW_TO_TRIGGER_AVATAR_EMOTE])
```
menduz, line 61: idiomatically `ensure` returns a sanitized or initializes the needed thing
and `require` or `assert` is the protocol assertion

PR #125, opened 2021-12-10
fix ci tests

gonpombo8:
# What? 
...

# Why? 
...


menduz:
What do you think if we make the `"printWidth"` property of eslint be overwritten by 120? the default 80 makes most `if` statements unreadable due to line breaks

leanmendoza (approved):

leanmendoza (approved):

PR #130, opened 2021-12-16
refactor allscene events.

gonpombo8:
Closes https://github.com/decentraland/sdk/issues/141

# What? 
Move allSceneEvents to a new saga folder.
...

# Why? 
To avoid having allSceneEvents sagas handler all over kernel. 
Add tests for scene events saga.
...


menduz:
🤩 yes

PR #140, opened 2021-12-20
feat: add a restricted websocket.

leanmendoza:
# What? 
Add a custom global `WebSocket` class to each scene. 
This is a first approach, let's discuss it.

# Why? 
It's neccesary the limitation according with Portable Experiences launch, to forbid inappropriate or abusively WebSocket usage.


packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -533,7 +534,28 @@ export abstract class SceneRuntime extends Script {
       }
 
       try {
-        await this.runCode(source as any as string, { dcl })
+        const self = this
+        class RestrictedWebSocket extends WebSocket {
```
leanmendoza, line 538: I'd like this isn't here, but I need the access to Permissions API, or maybe a promise which can access to it

gonpombo8: Maybe we can move this RestrictedWebScoket to a new file and create a fn to construct the permisisons. Something like



## SceneRuntime.ts
```ts
  const { Permissions } = await this.loadAPIs(['Permissions'])
  const restrictedWebSocket = createWebSocket(Permissions)
```
## WebSocket.ts
```ts
function createWebSocket(permission: Permissions) {
  return class RestrictedWebSocket extends WebSocket {
    constructor(url: string | URL, protocols?: string | string[]) {
      console.log({ url })
      if (url.toString().toLowerCase().substr(0, 4) !== 'wss:') {
...etc
```

or if you dont want to block the sceneRuntime you can pass a fn to createWebSocket

```ts
        async function getPermissions() {
          const { Permissions } = self.loadAPIs(['Permissions'])
          return Permissions
        }
```
And at `websocket.ts` call this async fn and then check if it has the valid permissions.
```ts
getPermissions().then(permission => permission.hasPermission....)
```

packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -533,7 +535,11 @@ export abstract class SceneRuntime extends Script {
       }
 
       try {
-        await this.runCode(source as any as string, { dcl })
+        const { Permissions } = await this.loadAPIs(['Permissions'])
+        const restrictedWebSocket = createWebSocket(Permissions)
+        const restrictedFetch = createFetch(Permissions)
+
+        await this.runCode(source as any as string, { dcl, WebSocket: restrictedWebSocket, fetch: restrictedFetch })
```
menduz, line 542: I'd also patch the globalThis.fetch, keeping the original fetch in a variable for internal usage inside the SceneRuntime

packages/scene-system/sdk/WebSocket.ts
```diff
@@ -0,0 +1,19 @@
+import { PermissionItem, Permissions } from 'shared/apis/Permissions'
+
+export function createWebSocket(permission: Permissions) {
+  return class RestrictedWebSocket extends WebSocket {
+    constructor(url: string | URL, protocols?: string | string[]) {
+      if (url.toString().toLowerCase().substr(0, 4) !== 'wss:') {
```
menduz, line 6: This MAY be enabled tho in preview mode. With a warning I pressume

packages/scene-system/sdk/Fetch.ts
```diff
@@ -0,0 +1,19 @@
+import { PermissionItem, Permissions } from 'shared/apis/Permissions'
+
+export type FetchFunction = (resource: RequestInfo, init?: RequestInit | undefined) => Promise<Response>
+
+export function createFetch(permission: Permissions, originalFetch: FetchFunction) {
+  return (resource: RequestInfo, init?: RequestInit | undefined): Promise<Response> => {
+    const url = resource instanceof Request ? resource.url : resource
+    if (url.toLowerCase().substr(0, 8) !== 'https://') {
```
menduz, line 8: This MAY be enabled tho in preview mode. With a warning I presume. It would be useful for development mode

packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -533,7 +538,30 @@ export abstract class SceneRuntime extends Script {
       }
 
       try {
-        await this.runCode(source as any as string, { dcl })
+        const { Permissions } = await this.loadAPIs(['Permissions'])
+
+        this.originalFetch = fetch
+        this.originalWebSocket = WebSocket
+
```
menduz, line 545: I was looking at the logic inside the createWebSocket and createFetch functions and there is a flaw in the permissions. It runs asynchronously and therefore don't fail properly to create the fetch.

To prevent that async race condition it may be good to fetch the permissions in this function and to pass the booleans to the new functions:

```suggestion
const canUseWebsocket = await permission.hasPermission(PermissionItem.USE_WEBSOCKET)
```

leanmendoza: I haven't got the flaw yet and the logic in the creation methods are different. 
# Websockets 
I don't really like the websocket way, because the constructor runs always and the permission stops its functionality after the check, so in the meanwhile time everyone has allow to use web sockets. But, I think we can trust the async permission getting is faster than websocket connection.

# Fetch
I thought this implementation it was okay because the overwritten function it's a wrapped promise which checks before the permission and if it's okay continue the regular functionality. 

Pros:
We can ask or give to devs the possibility of asking users for permissions. For example when you enable your camera and the browser asking you to enable it. I thought something similar. 
Checking the permission before the scene loads, through away this possibility

menduz (approved):

test/scene-system/sdk/usingFetchAndWebsocket.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+  globalThis.WebSocket = FakeWebSocket
+})
+
+after(() => {
+  globalThis.WebSocket = originalWebSocket
+  originalWebSocket = null
+})
+
+describe('Safe fetch and WebSocket permissions', () => {
+  it('fetch', async () => {
```
gonpombo8, line 29: its a good practice to split the tests so when we break something its easier to debug.
```ts
describe('Wrapped websocket', () => {
  it('should log an error if the ws is not secure', () => {
  
  })

  it('should run successfully if the ws is secure', () => {
  })
})
```
```ts
describe('Wrapped fetch', () => {
  it(...)
})

etc
```

gonpombo8 (approved):

menduz (dismissed):
Looks good, although haven't tested it

gonpombo8 (approved):

packages/scene-system/sdk/Fetch.ts
```diff
@@ -0,0 +1,27 @@
+export type FetchFunction = (resource: RequestInfo, init?: RequestInit | undefined) => Promise<Response>
+
+export interface FetchOptions {
+  canUseFetch: boolean
+  originalFetch: FetchFunction
+  previewMode: boolean
+  log: any
```
gonpombo8, line 7:
```ts
log(message: string): void
```

packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -103,6 +106,9 @@ export abstract class SceneRuntime extends Script {
 
   isPreview: boolean = false
 
+  private originalFetch: any
```
gonpombo8, line 109:
```suggestion
  private originalFetch!: typeof fetch
```

packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -103,6 +106,9 @@ export abstract class SceneRuntime extends Script {
...
+  private originalWebSocket: any
```
gonpombo8, line 110:
```suggestion
  private originalWebSocket!: typeof WebSocket
```

packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -533,7 +539,34 @@ export abstract class SceneRuntime extends Script {
...
+          canUseFetch,
+          originalFetch: this.originalFetch,
+          previewMode: this.isPreview || unsafeAllowed,
+          log: dcl.log
+        })
+
+        globalThis.fetch = restrictedFetch
+        globalThis.WebSocket = restrictedWebSocket
+
+        this.originalFetch
```
gonpombo8, line 566: whats this ? can we remove this two lines ?
```suggestion
```

packages/scene-system/sdk/WebSocket.ts
```diff
@@ -0,0 +1,24 @@
+export interface WebSocketClassOptions {
+  canUseWebsocket: boolean
+  previewMode: boolean
+  log: any
```
gonpombo8, line 4:
```ts
log(): void
```

PR #151, opened 2021-12-29
feat: implementing biw deployment

Suduck:
# What?

This PR gives the ability to deploy projects from builder in world in the lands


packages/shared/apis/SceneStateStorageController/utils.ts
```diff
@@ -1,6 +1,17 @@
...
-
+import { ContentClient } from 'dcl-catalyst-client'
+import { EntityType } from 'dcl-catalyst-commons'
+import { Asset, AssetId, BuilderAsset, CONTENT_PATH, DeploymentResult, PublishPayload } from './types'
+import { Authenticator } from 'dcl-crypto'
+import { getCurrentIdentity } from 'shared/session/selectors'
+import { createGameFile } from './SceneStateDefinitionCodeGenerator'
+import { ETHEREUM_NETWORK, BUILDER_SERVER_URL } from 'config'
+import { BASE_BUILDER_SERVER_URL_ROPSTEN } from './BuilderServerAPIManager'
+import { serializeSceneStateFromEntities } from 'scene-system/stateful-scene/SceneStateDefinitionSerializer'
```
menduz, line 13: Do not mix workers' code with main thread's code

packages/unity-interface/BrowserInterface.ts
```diff
@@ -664,6 +666,11 @@ export class BrowserInterface {
     getUnityInstance().SendHeaders(data.url, headers)
   }
 
+  public async PublishSceneState(data: PublishPayload) {
+    const result = await deployScene(data)
+    getUnityInstance().SendPublishSceneResult(result)
+  }
```
menduz, line 672: Prevent creation of async function in BrowserInterface, nobody catches the errors

```suggestion
  public PublishSceneState(data: PublishPayload) {
    deployScene(data).then(result => {
      getUnityInstance().SendPublishSceneResult(result)
    }).catch(error => defaultLogger.error(error))
  }
```

packages/shared/apis/SceneStateStorageController/utils.ts
```diff
@@ -16,6 +27,76 @@ const HUMAN_READABLE_TO_ID: Map<string, number> = new Map([
...
+      throw new Error('Identity not found when trying to deploy an entity')
+    }
+    const authChain = Authenticator.signPayload(identity, entityId)
+
+    // Deploy entity
+    await contentClient.deployEntity({ files, entityId, authChain })
+
+    result = { ok: true }
+  } catch (error) {
+    result = { ok: false, error: `${error}` }
```
menduz, line 86: Prevent state mutation, use plain-ol return statements

```suggestion
    return { ok: true }
  } catch (error) {
    return { ok: false, error: `${error}` }
```

menduz (approved):

PR #154, opened 2022-01-04
add unity-renderer cron job to verify the latest version installed

gonpombo8:
# What? 
...

# Why? 
...


menduz (approved):
let's give it a try

PR #159, opened 2022-01-07
fix: use content mapping instead of url for smart wearables icon

pbosio:
closes https://github.com/decentraland/sdk/issues/161

### DIFF
--- a/packages/shared/wearablesPortableExperience/sagas.ts
+++ b/packages/shared/wearablesPortableExperience/sagas.ts
@@ -126,8 +126,8 @@ function* handleStartWearablesPortableExperience(action: StartWearablesPortableE
       const mappings = wearableContent.map(($) => ({ file: getFile($.key), hash: $.hash }))
       const name = wearable.i18n[0].text
 
-      // TODO: make sure thumbnail is added to the mappings if not there to avoid creating an url for the icon here
-      const icon = baseUrl + wearable.thumbnail
+      const icon = "smartWearableIcon"
+      mappings.push({file: icon, hash: wearable.thumbnail})
 
       spawnPortableExperience(wearable.id, 'main', name, baseUrl, mappings, icon).catch((e) => defaultLogger.error(e))
     } catch (e) {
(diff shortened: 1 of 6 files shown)


packages/shared/wearablesPortableExperience/sagas.ts
```diff
@@ -126,8 +126,8 @@ function* handleStartWearablesPortableExperience(action: StartWearablesPortableE
       const mappings = wearableContent.map(($) => ({ file: getFile($.key), hash: $.hash }))
       const name = wearable.i18n[0].text
 
-      // TODO: make sure thumbnail is added to the mappings if not there to avoid creating an url for the icon here
-      const icon = baseUrl + wearable.thumbnail
+      const icon = 'smartWearableIcon'
+      mappings.push({ file: icon, hash: wearable.thumbnail })
```
menduz, line 130: @pbosio @leanmendoza how is this supposed to work?

pbosio: `wearable.thumbnail` comes with the hash from the content server
`const icon = 'smartWearableIcon'` it's just to add a name for that file
`mappings.push(...)` is for adding it to the content mapping of the portable experience.
It would reach renderer here with the key-value `'smartWearableIcon':'SomeHash'`
https://github.com/decentraland/unity-renderer/blob/9e1d704335104dd9ee266d782793122d35a27e5a/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs#L793
where renderer will check if `smartWearableIcon` exists in the mappings and build the url for the corresponding hash

leanmendoza: I just tested it this branch and is already working, so we only have to change the `'smartWearableIcon'` to `'menuBarIcon'` right?

leanmendoza: Because it's broken right now

packages/shared/wearablesPortableExperience/sagas.ts
```diff
@@ -126,8 +126,8 @@ function* handleStartWearablesPortableExperience(action: StartWearablesPortableE
       const mappings = wearableContent.map(($) => ({ file: getFile($.key), hash: $.hash }))
       const name = wearable.i18n[0].text
 
-      // TODO: make sure thumbnail is added to the mappings if not there to avoid creating an url for the icon here
-      const icon = baseUrl + wearable.thumbnail
+      const icon = 'smartWearableMenuBarIcon'
```
leanmendoza, line 129: Yesterday we tested with 'smartWearableIcon', is this already at unity-side?

leanmendoza (approved):

PR #163, opened 2022-01-10
Add filter by capacity to catalysts

jmoguilevsky:
# What? 
This pr prevents selecting a catalyst that has a max concurrent users lower or equal to the user count

Removes the zip which depended on the size of both arrays, now it's one map that takes care of building the correct information before filtering

# Why? 
When a catalyst is not able to accept new connections and an algorithm still picks it, the user experiences an error that's not solvable without refreshing, and in some cases, kernel was still selecting the same realm (for example when the algorithm is 100% load balancing)

### DIFF
--- a/packages/shared/dao/index.ts
+++ b/packages/shared/dao/index.ts
@@ -17,23 +17,22 @@ import {
   getCatalystRealmCommsStatus,
   getRealm,
   getAllCatalystCandidates,
-  areCandidatesFetched,
-} from "./selectors"
-import { fetchCatalystNodesFromDAO } from "shared/web3"
-import { setCatalystRealm, setCatalystCandidates } from "./actions"
-import { deepEqual } from "atomicHelpers/deepEqual"
-import { worldToGrid } from "atomicHelpers/parcelScenePositions"
-import { lastPlayerPosition } from "shared/world/positionThings"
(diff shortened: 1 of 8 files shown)


gonpombo8 (dismissed):
can we add some unit tests for this fn ?

packages/shared/dao/index.ts
```diff
@@ -156,70 +155,72 @@ export function commsStatusUrl(domain: string, includeLayers: boolean = false, i
...
-          status: status!,
-          elapsed: elapsed!,
-          lighthouseVersion: result!.version,
-          catalystVersion: result!.env.catalystVersion
-        }
+  const results = (
+    await Promise.all(
+      nodes.map(async (node) => ({ domain: node.domain, ...(await ping(commsStatusUrl(node.domain, true, true))) }))
+    )
+  ).filter((realm) => (realm.result?.maxUsers ?? 0) > (realm.result?.usersCount ?? -1))
```
gonpombo8, line 177: we could add this filter inside the `.reduce` fn, like we do when `status !== ServerConnectionStatus.OK`

if (condition) { return union }

jmoguilevsky: I'd say having the filter before the reduce is much simpler to read, but i can add it in the reduce

gonpombo8 (approved):

gonpombo8 (approved):

PR #168, opened 2022-01-13
fix: report a wrong position after movePlayerTo is called.

leanmendoza:
# What? 
Change `newPosition` to `position`. NewPosition is the function argument, and position is the finally calculated position.

# Why? 
Fix https://github.com/decentraland/sdk/issues/162


gonpombo8 (commented):
can we change newPosition to `pos` or something less confusing ?
For future devs. 😄

gonpombo8 (approved):
thanks :✨

PR #171, opened 2022-01-17
fix: problem with catalyst helper function

menduz:
# What? 
Fixes a problem by deleting it

# Why? 
...

### DIFF
--- a/packages/decentraland-loader/lifecycle/controllers/parcel.ts
+++ b/packages/decentraland-loader/lifecycle/controllers/parcel.ts
@@ -32,7 +32,7 @@ export class ParcelLifeCycleController extends EventEmitter {
   public resetScenesOnSight() {
     this.currentlySightedParcels.clear()
 
-    this.parcelStatus.forEach((value: ParcelLifeCycleStatus, key: string) => {
+    this.parcelStatus.forEach((value: ParcelLifeCycleStatus, _key: string) => {
       value.setOffSight()
     })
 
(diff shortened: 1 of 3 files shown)


pbosio (approved):

packages/decentraland-loader/lifecycle/controllers/parcel.ts
```diff
@@ -32,7 +32,7 @@ export class ParcelLifeCycleController extends EventEmitter {
   public resetScenesOnSight() {
     this.currentlySightedParcels.clear()
 
-    this.parcelStatus.forEach((value: ParcelLifeCycleStatus, key: string) => {
+    this.parcelStatus.forEach((value: ParcelLifeCycleStatus, _key: string) => {
```
gonpombo8, line 35: thanku

gonpombo8 (approved):

menduz:
I pushed a change to reflect ADR53

Also, gon. In regards to catalyst client: I believe we are currently using only two functions that can be replaced with a plaing `fetch`. You may find interesting that we can delete the catalyst-client usage from this repo @gonpombo8

gonpombo8:
@menduz nice. Issue created https://github.com/decentraland/sdk/issues/169 🎉

gonpombo8 (approved):

menduz:
@nachomazzara @pbosio we can use a second validation before merging please?

pbosio:
> @nachomazzara @pbosio we can use a second validation before merging please?

It's working for me. The only thing worth mention is that with this changes, if your are in a test catalyst and you want to test wearables deployed in org alongside with wearables in testing catalyst you can't.
But I think that's not a common scenario at all... :)

PR #187, opened 2022-02-04
feat: Add SetDisabledPortableExperiences into BrowserInterface

sandrade-dcl:
# What? 
We have added the `SetDisabledPortableExperiences` function into `BrowserInterface` in order to be able to disable Portable Experiences from the renderer.

# Why? 
It is needed to be used by the new Experiences Viewer that we're implementing in the Unity side (https://github.com/decentraland/unity-renderer/pull/1873).


menduz (changes requested):
Adding this functionality gives so much power to the renderer enabling possible inconsistencies and exploitable bugs.

As an example, there are no way to prevent the experience from reloading after you change the nick of your avatar or backpack.

Please let's organize a call to evaluate which may be the possible solution for this use case from the runtime/kernel.

I'm thinking about specifying which portable experiences are turned off. Adopting a "normal open" approach.

```ts
function SetDisabledPortableExperiences(args: { urns: string[] }): void
```

olavra:
Triggering pex from unity should be something feasible from an user stand point. In fact, automatically launching the pex when you equip the wearable instead of being an user choice is not the best way imo. 
In the future, POAPs or Items may trigger pex from the inventory, but as an user choice triggering a button, not because its equipped in the avatar, or we come with a ghost equipping system.
I think this current situation is an special case while we trigger PEX associated to wearable equipment, and right now is a bad user experiencie, as enabling a previously disabled pex, requires to unequip (replacing by another item that uses the same spot), going to the backpack and equipping the item again.

sandrade-dcl:
In order to add more context to this discussion... As far as I know, according to the requirements already discussed and validated from the product side, we need to be able to start a PEX from the new HUD we are implementing:

image

**Figma design**: https://www.figma.com/file/BmxUcVf2J3NhcXvlZcLHwt/PEX

nearnshaw:
For context:
The list of Experiences listed in that UI should only reflect the experiences linked to the wearables you currently have on.  It wouldn't allow you to turn on ANY portable experience.

Having a wearable on shouldn't necessarily map to having its experience running.

Another point that is probably related to this too:

Players should be able to reject turning on a portable experience if when they're prompted with the list of permissions that the portable experience requires they decline.
https://www.notion.so/decentraland/RFC-Portable-experience-controlled-actions-fde77221063143fd9697aa7d69612f4d

menduz:
In that case, all of the required conditions are sufficed stating which portable experiences are supposed to be turned off.

```ts

// call this function with all disabled portable experiences that should not run when the backpack
// or poaps or any other condition changes
function SetDisabledPortableExperiences(args: { urns: string[] }): void 
```

Continues being the recommended approach. The renderer should never know about the internals of portable experiences or other runtime related matters. That would increase the overlap and coupling between teams by far and making both projects more co-dependent.

Kernel works with declarative code, not imperative. That means we shoudn't tell kernel what to do, instead, we should tell kernel what do we expect. In this case, kernel controls the resources and lifecycles of PX. We should tell kernel which PX should _not run_ to allow it to gracefully control the lifecycle of PX and preventing code duplication and spread responsibilities.

packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -203,3 +214,20 @@ export async function spawnPortableExperience(
...
+  idsToDisable.forEach(async (pexId) => {
+    if (currentPortableExperiences.has(pexId)) {
+      await killPortableExperienceScene(pexId)
+    }
+  })
+
+  const profile = getCurrentUserProfile(store.getState())
+  profile?.avatar.wearables.forEach((wearableId) => {
+    if (!idsToDisable.includes(wearableId)) {
+      // TODO: SPAWN PORTABLE EXPERIENCE WITH ID = wearableId (if it is an equipped smart wearable)
```
sandrade-dcl, line 228: Hey @menduz could you give me a hand with this `TODO`? I need to get the equipped wearables info in order to be able to call the `spawnPortableExperience` function ans so start again the non-ignored PEXs. Thanks!

menduz:
Santi this should be done using sagas. For "urgency" reasons I'll start it to unblock the issue

sandrade-dcl:
> Santi this should be done using sagas. For "urgency" reasons I'll start it to unblock the issue

Thanks @menduz!

sandrade-dcl:
@menduz the flow is already working as we expected. The only thing that this implementation is missing is to also support the launching/stopping of PEXs directly from browser console (this would be only for `local` and `.zone`), for example:
```
spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:radio", "main")
```
Currently, when we launch PEXs in this way, after any profile update, all these PEXs are being stopped because it seems they didn't were into account as "desirable" PEXs in the same way that the wearables-related so.

packages/shared/apis/PortableExperiences.ts
```diff
@@ -34,10 +40,10 @@ export class PortableExperiences extends ExposableAPI {
   @exposeMethod
   async kill(pid: PortableExperienceUrn): Promise<boolean> {
     const parcelIdentity: ParcelIdentity = this.options.getAPIInstance(ParcelIdentity)
-    const portableExperience: PortableExperienceHandle | undefined = await getPortableExperience(pid)
+    const portableExperience = getRunningPortableExperience(pid)
 
     if (!!portableExperience && portableExperience.parentCid === parcelIdentity.cid) {
-      return await killPortableExperienceScene(pid)
+      store.dispatch(removeDebugPortableExperience(pid))
```
gonpombo8, line 46: kill will always return false since we remove the `return` key from here. Is that expected ?

gonpombo8 (approved):

packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -3,68 +3,63 @@ import {
...
+  store.dispatch(removeDebugPortableExperience(urn))
 }
 
-export async function killPortableExperienceScene(sceneUrn: string): Promise<boolean> {
+function killPortableExperienceScene(sceneUrn: string) {
   const peWorker = getSceneWorkerBySceneID(sceneUrn)
   if (peWorker) {
     forceStopParcelSceneWorker(peWorker)
     currentPortableExperiences.delete(sceneUrn)
+    store.dispatch(removeDebugPortableExperience(sceneUrn))
```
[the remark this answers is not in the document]
sandrade-dcl: This is what I had to add to fix the problem we had with the "debug" PEXs. If we don't call to `removeDebugPortableExperience` here, we cannot start a "debug" PEX after already having started/stopped it the first time.

The steps to reproduce this issue were:
1. I spawn a debug PEX from the browser console:
```
spawnDebugPortableExperienceSceneFromUrn("urn:decentraland:off-chain:static-portable-experiences:radio", "main")
```
2. The "radio" PEX is executed correctly.
3. Now, from the renderer, we make the next calls in a row:
```
SetDisabledPortableExperiences <- passing deny list WITH the debug PEX id included, in order to stop it.
SetDisabledPortableExperiences <- passing the deny list WITHOUT the debug PEX id included, in order to let the system to start it again when we want.
```

In this scenario, if we don't make the call to `store.dispatch(removeDebugPortableExperience(sceneUrn))`, the debug PEX never starts again.

packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -50,7 +50,6 @@ function killPortableExperienceScene(sceneUrn: string) {
   if (peWorker) {
     forceStopParcelSceneWorker(peWorker)
     currentPortableExperiences.delete(sceneUrn)
-    store.dispatch(removeDebugPortableExperience(sceneUrn))
```
sandrade-dcl: @menduz this change is provoking the issue again:
1. I start a PEX from browser console:
```
spawnDebugPortableExperienceSceneFromUrn("urn:decentraland:off-chain:static-portable-experiences:portal-experience", "main")
```
2. The PEX is executed correctly.
3. I stop the PEX from Renderer doing the next calls to Kernel:
```
SetDisabledPortableExperiences <-- Passing a list with the PEX id. It will stop the PEX, and keep it as "denied".
SetDisabledPortableExperiences <-- Passing a list without the PEX id. It will extract the PEX from the list of "denied".
```
4. The PEX scene is stopped correctly.
5. I execute the step 1 again.
6. The PEX is no longer executed.

sandrade-dcl (commented):
The issue is happening again 😭

leanmendoza (dismissed):

pbosio (approved):

menduz (approved):

PR #188, opened 2022-02-06
depreciated peer parameter

RishiKumarRay:
# What? 
Remove or comment code for /peer-parameters in kernel


leanmendoza:
Hey @RishiKumarRay , thx for the PR, validations have failed. Don't forget to compile successfully your code before pushing

leanmendoza:
> Screenshot from 2022-02-07 19-29-56 can you help me with these errors ? @leanmendoza

Hey @RishiKumarRay , will you continue with this PR?

The errors are because you partially removed function and there is some code in top-level. Please fix the code removing all the original function that you already removed.

A good practice is to run the `make watch` while you're developing, so you can see if the code it's broken or not.

PR #192, opened 2022-02-09
add console wrapper

gonpombo8:
# What? 
Add a string prefix to the logs so we can identify if the logs comes from kernel or another library.
This way we can use the DEBUG_PREFIX url param to filter the logs that we want.

i.e. https://play.decentraland.zone/?kernel-branch=add/wrap-console-log&DEBUG_PREFIX=kernel: will log only the logs that starts with `kernel:`

By default we log everything, so this should not change anything if we dont use the query param.
...

# Why? 
Clean the noise in the console so the user can see only the logs that he wants. i.e. scene logs
...

### DIFF
--- a/packages/config/index.ts
+++ b/packages/config/index.ts
@@ -132,6 +132,7 @@ export const DEBUG_LOGIN = location.search.includes('DEBUG_LOGIN')
 export const DEBUG_PM = location.search.includes('DEBUG_PM')
 export const DEBUG_SCENE_LOG = DEBUG || location.search.includes('DEBUG_SCENE_LOG')
 export const DEBUG_KERNEL_LOG = !PREVIEW || location.search.includes('DEBUG_KERNEL_LOG')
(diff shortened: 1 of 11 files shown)


menduz:
Can you expand on the WHY? The logs are really valuable to hide them. Plus https://owasp.org/www-community/vulnerabilities/Poor_Logging_Practice

gonpombo8:
> Can you expand on the WHY? The logs are really valuable to hide them. Plus https://owasp.org/www-community/vulnerabilities/Poor_Logging_Practice https://owasp.org/www-project-top-ten/2017/A10_2017-Insufficient_Logging%2526Monitoring https://cwe.mitre.org/data/definitions/778.html

We are not removing any logs. Its just adding some prefix so we can hide/show them as we wish. So it doesn't change anything unless u use the debug_prefix .
Enter to play.decentraland.org and try to find some log for debugging. It's almost impossible. 
The logs are valuable, but when u log util things.

PR #196, opened 2022-02-14
Chore/remove catalyst client dependency

doomling:
# What? 
...

# Why? 
...

### DIFF
--- a/packages/decentraland-loader/lifecycle/controllers/download.ts
+++ b/packages/decentraland-loader/lifecycle/controllers/download.ts
@@ -1,186 +1,180 @@
-import { future, IFuture } from "fp-future";
-import { ILand } from "shared/types";
-import { EmptyParcelController } from "./EmptyParcelController";
-import { WorldConfig } from "shared/meta/types";
-import fetch from "node-fetch";
+import { future, IFuture } from 'fp-future'
+import { ILand } from 'shared/types'
+import { EmptyParcelController } from './EmptyParcelController'
+import { WorldConfig } from 'shared/meta/types'
+import fetch from 'node-fetch'
 
 export type DeployedScene = {
-  parcel_id: string;
-  root_cid: string;
-  scene_cid: "";
-};
-
-export type SceneMappingResponse = {
-  data: Array<DeployedScene>;
-};
-
-function getSceneIdFromSceneMappingResponse(scene: DeployedScene) {
-  return scene.root_cid;
+  parcel_id: string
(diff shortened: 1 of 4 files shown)


packages/decentraland-loader/lifecycle/controllers/download.ts
```diff
@@ -1,9 +1,8 @@
 import { future, IFuture } from 'fp-future'
 import { ILand } from 'shared/types'
-import { CatalystClient } from 'dcl-catalyst-client'
-import { EntityType } from 'dcl-catalyst-commons'
 import { EmptyParcelController } from './EmptyParcelController'
 import { WorldConfig } from 'shared/meta/types'
+import fetch from 'node-fetch'
```
menduz, line 5: the global `fetch` should be used in the browser

packages/shared/catalogs/sagas.ts
```diff
@@ -100,7 +105,10 @@ function* fetchWearablesFromCatalyst(filters: WearablesRequestFilters) {
       // Fetch published collections
       const urnCollections = collectionIds.filter((collectionId) => collectionId.startsWith('urn'))
       if (urnCollections.length > 0) {
-        const zoneWearables: PartialWearableV2[] = yield client.fetchWearables({ collectionIds: urnCollections })
+        const zoneWearables: PartialWearableV2[] = yield jsonFetch(
```
menduz, line 108:
```suggestion
        const zoneWearables: PartialWearableV2[] = yield call(jsonFetch,
```

What would happen in case of failure in the response?

packages/shared/catalogs/sagas.ts
```diff
@@ -131,7 +139,15 @@ function* fetchWearablesFromCatalyst(filters: WearablesRequestFilters) {
       }
     }
   } else {
-    const wearables: PartialWearableV2[] = yield call(fetchWearablesByFilters, filters, client)
+    // Should I migrate the whole query logic to avoid the dependency?
+    // See logic here splitValuesIntoManyQueries -> /kernel/node_modules/dcl-catalyst-client/dist/utils/Helper.js
```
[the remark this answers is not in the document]
agusaldasoro: For the endpoint `/lambdas/collection/wearables` you can iterate between pages using: 
`lastId` https://decentraland.github.io/catalyst-api-specs/#operation/getWearables

In the API Specs should be all the filters/pagination query parameters needed, if something is not matching the reality or needs a change, please let us know!

PR #206, opened 2022-02-17
feat: invalidate scenes when publishing

Suduck:
# What?

When you are publishing in builder in world if the publish goes well, it will invalidate the data of all the scenes

# Why? 

We want to see the changes that you have published when you exit from the editor, if we don't invalidate the scenes we are not able to see them


packages/unity-interface/BrowserInterface.ts
```diff
@@ -682,6 +682,7 @@ export class BrowserInterface {
     deployScene(data)
       .then(() => {
         deploymentResult = { ok: true }
+        invalidateAllScenes()
```
gonpombo8, line 685:
```suggestion
        void invalidateAllScenes()
```

packages/decentraland-loader/lifecycle/manager.ts
```diff
@@ -108,6 +108,12 @@ export class LifecycleManager extends TransportBasedServer {
     }
   }
 
+  async invalidateAllScenes() {
+    for (const entry of Array.from(this.sceneIdToRequest.entries())) {
+      await this.invalidateScene(entry[0])
```
gonpombo8, line 113:
```suggestion
    for (const sceneId of this.sceneIdToRequest.keys()) {
      await this.invalidateScene(sceneId)
```

packages/unity-interface/BrowserInterface.ts
```diff
@@ -686,6 +686,8 @@ export class BrowserInterface {
     deployScene(data)
       .then(() => {
         deploymentResult = { ok: true }
+        if (!data.reloadSingleScene) void invalidateAllScenes()
```
menduz, line 689: Please revert the order of the if and remove the ! for clarity

menduz: In which cases is `reloadSingleScene=false` needed?

Suduck: I will! 

It is needed when you are publishing inside the editor, if you are inside the editor you don't need to reload a single scene since they will be loaded when you exit, however if you are publishing outside the editor ( from the panel) you will need to reload that scene

packages/shared/world/parcelSceneManager.ts
```diff
@@ -186,6 +186,63 @@ export const parcelSceneLoadingState: ParcelSceneLoadingState = {
   isolatedModeOptions: null as IsolatedModeOptions | null
 }
 
+async function isEmptyParcel(coord: string): Promise<boolean> {
+  if (!parcelSceneLoadingState.lifecycleManager) return false
+
+  for (const record of parcelSceneLoadingState.lifecycleManager.sceneIdToRequest) {
```
pbosio, line 192: Is it possible to generate the empty scene id and just check if `parcelSceneLoadingState.lifecycleManager.sceneIdToRequest.get(emptySceneId)` is undefined?

packages/decentraland-loader/lifecycle/utils/reloadScene.ts
```diff
@@ -5,3 +5,9 @@ export async function reloadScene(sceneId: string) {
   if (!server) return
   return server.reloadScene(sceneId)
 }
+
+export function invalidateAllScenes(coordsToInvalidate: string[]) {
+  const server = getServer()
+  if (!server) return
+  return server.invalidateAllScenes(coordsToInvalidate)
+}
```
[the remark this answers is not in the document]
Suduck: Moved to invalidateScenes and renamed for more clarity

packages/decentraland-loader/lifecycle/worker.ts
```diff
@@ -136,6 +136,10 @@ let downloadManager: SceneDataDownloadManager
       connector.on('Scene.Invalidate', (data: { sceneId: string }) => {
         sceneController.invalidate(data.sceneId)
       })
+
+      connector.on('Scene.InvalidateByCoords', (data: { coords: string[] }) => {
```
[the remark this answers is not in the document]
Suduck: Done

packages/decentraland-loader/lifecycle/manager.ts
```diff
@@ -108,7 +108,25 @@ export class LifecycleManager extends TransportBasedServer {
     }
   }
 
+  async invalidateAllScenes(coordsToInvalidate: string[] | undefined) {
+    for (const sceneId of this.sceneIdToRequest.keys()) {
+      await this.invalidateSceneAndCoords(sceneId)
+    }
+    if (coordsToInvalidate) this.notify('Scene.InvalidateByCoords', { coords: coordsToInvalidate })
+  }
```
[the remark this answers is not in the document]
Suduck: This method will invalidate all the scenes and if you pass it coordinates, it will also invalidate the coordinates for the parcels. That is because if we publish an scene of 2x2 in an scene of 1x1 plus 3 empty parcels, we need to invalidate not only the scenes, but also the parcel in order to download the new scene

packages/shared/world/parcelSceneManager.ts
```diff
@@ -372,6 +429,20 @@ export async function loadParcelSceneByIdIfMissing(sceneId: string) {
   }, WORKER_TIMEOUT)
 }
 
+function removeDesiredParcel(sceneId: string) {
+  const desiredScenes = getDesiredParcelScenes()
+  if (!desiredScenes.has(sceneId)) return
```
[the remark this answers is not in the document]
Suduck: done

packages/shared/world/parcelSceneManager.ts
Suduck: No, I don't remember exactly why but I remember that we needed to check the base position of the scene because it could change of size when we invalidate it

pbosio (approved):

PR #212, opened 2022-02-22
chore: improved show fps command

davidejensen:
As requested by @olavra and then stated in the unity-renderer issue https://github.com/decentraland/unity-renderer/issues/1950 this PR renames the previous chat command /showfps to /debug and it adds the versioning in the chat command /debug


packages/shared/chat/sagas.ts
```diff
@@ -289,9 +289,14 @@ function initChatCommands() {
     }
   })
 
-  addChatCommand('showfps', 'Show FPS counter', (_message) => {
+  addChatCommand('debug', 'Show debug panel', (_message) => {
```
menduz, line 292: Can we keep both? People have muscular memory for showfps

packages/shared/chat/sagas.ts
```diff
@@ -289,9 +289,14 @@ function initChatCommands() {
...
     fpsConfiguration.visible = !fpsConfiguration.visible
     fpsConfiguration.visible ? getUnityInstance().ShowFPSPanel() : getUnityInstance().HideFPSPanel()
+    const versions = injectVersions({})
+    const kernelVersion = versions['@dcl/kernel'] || 'unknown'
+    const rendererVersion = versions['@dcl/unity-renderer'] || 'unknown'
+
+    getUnityInstance().SetLocalDCLVersion(`\nKernel: ${kernelVersion}\nRenderer: ${rendererVersion}`)
```
menduz, line 299: This is being executed in several places. Are you sure there are no other (current) function doing the same?

packages/unity-interface/IUnityInterface.ts
```diff
@@ -98,6 +98,7 @@ export interface IUnityInterface {
   /** @deprecated send it with the kernelConfigForRenderer instead. */
   SetSceneDebugPanel(): void
   ShowFPSPanel(): void
+  SetLocalDCLVersion(version: string): void
```
menduz, line 101: This PR should also contain the newer unity-renderer version including that extra function otherwise the merge will not work. Also consider that the release of unity-renderer rollout should be at 100% before merging this Pr

menduz (commented):
Thanks for contributing to kernel! A couple of comments:

davidejensen:
> Thanks for contributing to kernel! A couple of comments:

Sorry I completely missed the message! I made the issue mainly to have the code committed but I have a few questions regarding the changes on the kernel and additional things I might need to add, I'll reply on the comments now :)

packages/shared/chat/sagas.ts
davidejensen: I don't know if it's a good idea to have both because it means 2 functions that do the same thing, so there is more "noise" in the chat /help command. Debug in my opinion makes more sense because it's a bit more descriptive, but if people are used to the /showfps we can keep it like that. The change was requested by @olavra let's see what he thinks as well

packages/shared/chat/sagas.ts
davidejensen: For what I've seen there is not a function doing this atm, I'll re check and in case create one

packages/unity-interface/IUnityInterface.ts
davidejensen: Yeah this is one of the things I wanted to ask, I'm still quite ignorant regarding how the kernel works and what is the correct workflow when pushing changes on kernel/unity-render that require the other

davidejensen:
To test the changes together with the unity branch: https://play.decentraland.org/?island=Inru&kernel-branch=chore%2Frename-fps-command&position=0%2C-1&realm=dg&renderer-branch=chore%2Fdebug-panel-improvs

packages/shared/chat/sagas.ts
menduz: Sorry this is a no-go, we should keep the previous one as well

davidejensen: Alright :) in this case if I can, I'd keep just the old one and avoid having the /debug one to avoid confusion

davidejensen:
unity-renderer PR is in revision: https://github.com/decentraland/unity-renderer/pull/1962

packages/shared/chat/sagas.ts
```diff
@@ -285,22 +285,13 @@ function initChatCommands() {
...
 
-    return {
-      messageId: uuid(),
-      sender: 'Decentraland',
-      messageType: ChatMessageType.SYSTEM,
-      timestamp: Date.now(),
-      body: 'Toggling FPS counter'
-    }
-  })
+  addChatCommand('showfps', 'Show fps panel', (_message) => getDebugPanelMessage())
```
eordano, line 294:
```suggestion
  addChatCommand('showfps', 'Show fps panel (deprecated in favor of /debug)', (_message) => getDebugPanelMessage())
```

packages/shared/chat/sagas.ts
```diff
@@ -285,22 +285,13 @@ function initChatCommands() {
       messageType: ChatMessageType.SYSTEM,
       sender: 'Decentraland',
       timestamp: Date.now(),
-      body: strings ? `Players around you:\n${strings}` : 'No other players are near to your location'
+      body: strings ? `Players around you:\n ${strings}` : 'No other players are near to your location'
```
eordano, line 288:
```suggestion
      body: strings ? `Players around you:\n${strings}` : 'No other players are near to your location'
```

packages/shared/chat/sagas.ts
eordano: I don't feel strongly about such a long message, maybe we could hide it from the `/help` list in some other way?

eordano (approved):

packages/shared/chat/sagas.ts
davidejensen: Solved with latest commit

packages/shared/chat/sagas.ts
davidejensen: We have longer messages in the /help chat, currently it appears like this:

Screenshot 2022-03-07 at 11 26 10

davidejensen:
@menduz  @pbosio after talking with @olavra we decided to avoid all the changes but the one of the new command, i changed the code to include just that small change with the new command.

gonpombo8 (approved):

PR #217, opened 2022-03-01
feat: only show request web3 api use on require sign

kuruk-mm:
The pop up of the transaction is been sent even if not needed.

We must add a list of the methods that need user validation to fix it properly.


leanmendoza (approved):

kuruk-mm:
Custom protocol test url: https://renderer-artifacts.decentraland.org/desktop/index.html?dcl://kernel-branch=fix/send-async-no-prompt

PR #221, opened 2022-03-02
Add/fetch timeout

gonpombo8:
# What? 
Add a timeout opt for fetch requests
...

# Why? 
We currently have one concurrent fetch per scene limitation, so when a request takes too long, then it blocks all the upcoming. Adding a { timeout: number } opt to the fetch so u can override the browser default timeout ( 300s ).

Also i changed the default timeout to 30000ms => 30s. (We can discuss about this number if u want)

```ts
fetch('https://some-url.casla', {}, { timeout: 100 }) 
...

### DIFF
--- a/packages/scene-system/sdk/Fetch.ts
+++ b/packages/scene-system/sdk/Fetch.ts
@@ -30,13 +30,11 @@ export function createFetch({ canUseFetch, previewMode, log, originalFetch }: Fe
 
     async function fetchRequest() {
       const abortController = new AbortController()
-      const signal = abortController.signal
-      const TIMEOUT_LIMIT = 5000
+      const TIMEOUT_LIMIT = 30000
       const timeout = setTimeout(() => {
         abortController.abort()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by leanmendoza — APPROVED** [review]
LGTM! Nice tests :D

File: packages/scene-system/sdk/Fetch.ts [hunk 1]
```diff
@@ -8,9 +8,13 @@ export interface FetchOptions {
...
 }
 
+type Opts = {
+  timeout?: number
+}
+
 export function createFetch({ canUseFetch, previewMode, log, originalFetch }: FetchOptions) {
   const fifoFetch = new PQueue({ concurrency: 1 })
-  return (resource: RequestInfo, init?: RequestInit | undefined): Promise<Response> => {
+  return (resource: RequestInfo, init?: RequestInit | undefined, opts?: Opts): Promise<Response> => {
```
menduz, line 17: @gonpombo8 before releasing a stable version, may I suggest to add timeout to the second argument instead adding an extra arg?

PR #242, opened 2022-03-22
feat: emotes command deprecation warning

AjimenezDCL:
Emotes commands will be deprecated soon, in the meantime we show a warning when using a `/emote {emoteId}` command.

The warning will only appear if the feature flag `emotes_customization` is on.


packages/shared/chat/sagas.ts
```diff
@@ -325,12 +327,17 @@ function initChatCommands() {
 
       getUnityInstance().TriggerSelfUserExpression(expression)
 
+      let body = ''
+      if (isFeatureEnabled(store.getState(), FeatureFlags.EMOTES_CUSTOMIZATION, false))
+        body =
+          '[WARNING] This command will be deprecated soon, take a look at our brand new emotes wheel by pressing B!'
```
menduz, line 333: Why are we deprecating this? Operative systems create their UI but never remove the command line `cp` commands

AjimenezDCL: It's a product requirement.

Now users can equip up to 10 emotes on their profile, this allows us to control the memory since we dont have to load all the GLTFs containing the animations for people with hundreds of NFT Emotes. 
The command feature brings too little compared to the confussion caused by the fact that you can type `/emote wave` while not having the Wave emote equipped and thus nothing happening.

PR #250, opened 2022-03-29
add timeout opts to signed fetch

gonpombo8:
# What? 
Add default timeout for singedFetch requests.
...

# Why? 
We add this functionality to all the scene requests but forgot to add it to the signedFetch.
signedFetch use the same type as fetch so the timeout opt was already added but the implementation was missing 😢 
...

### DIFF
--- a/packages/scene-system/sdk/Fetch.ts
+++ b/packages/scene-system/sdk/Fetch.ts
@@ -10,7 +10,7 @@ export interface FetchOptions {
   log(...a: any[]): void
 }
 
-type Opts = {
+export type Opts = {
   timeout: number
 }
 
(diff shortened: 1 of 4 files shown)


packages/scene-system/sdk/Fetch.ts
```diff
@@ -33,16 +33,20 @@ export function createFetch({ canUseFetch, previewMode, log, originalFetch }: Fe
...
-      clearTimeout(timeout)
-      return response
-    }
-
-    return fifoFetch.add(fetchRequest)
+    return fifoFetch.add(() => fetchWithTimeout(originalFetch, resource, init))
   }
 }
+
+export async function fetchWithTimeout(
```
leanmendoza, line 40: can we move this function outside of scene-system? scene-system is another project and could be external repo and atomicsHelpers shouldn't depend on it

leanmendoza (approved):

menduz (approved):

packages/scene-system/sdk/Fetch.ts
menduz: @gonpombo8

menduz:
Already in master

PR #265, opened 2022-04-11
feat: add report log from unity

kuruk-mm:
Fix: decentraland/explorer-desktop#173

### DIFF
--- a/packages/unity-interface/BrowserInterface.ts
+++ b/packages/unity-interface/BrowserInterface.ts
@@ -64,7 +64,7 @@ import { WearablesRequestFilters } from 'shared/catalogs/types'
 import { fetchENSOwnerProfile } from './fetchENSOwnerProfile'
 import { ProfileAsPromise } from 'shared/profiles/ProfileAsPromise'
 import { profileToRendererFormat } from 'shared/profiles/transformations/profileToRendererFormat'
-import { AVATAR_LOADING_ERROR, renderingActivated, renderingDectivated } from 'shared/loading/types'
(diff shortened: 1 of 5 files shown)


packages/unity-interface/BrowserInterface.ts
```diff
@@ -769,6 +769,26 @@ export class BrowserInterface {
   public ReportDecentralandTime(data: any) {
     setDecentralandTime(data)
   }
+
+  public ReportLog(data: { type: string; message: string }) {
+    switch (data.type) {
+      case 'trace':
+        defaultLogger.trace(data.message)
```
menduz, line 776: please create a unityLogger for this 🙏

packages/unity-interface/UnityInterface.ts
```diff
@@ -62,7 +62,33 @@ function resizeCanvas(targetHeight: number) {
   }
 }
 
+export function createUnityLogger(): ILogger {
```
gonpombo8, line 65: whats the diff between this fn and https://github.com/decentraland/kernel/blob/main/packages/shared/logger/index.ts#L29 ?

If the prefix its the only thing that changes, could we refactor that fn and use it for both cases ?

kuruk-mm: The first prefix is `kernel`, and I changed it to `unity`.

I can modify the first function to get that. But I created a new function because if we want to modify something to report the logs from Unity, for example, not showing normal logs, or we want to send directly to rollbar and not show those in console, we can.

This will add A LOT of messages.

If you want, I can modify the first function to have a new parameter to change that `kernel` prefix. And when we need what I said, we create a new function like this one.

gonpombo8 (approved):

PR #267, opened 2022-04-12
feat: add tracking to friends events

AjimenezDCL:
Added tracking to friendship updates events.

Added MessageType to chat events.

### DIFF
--- a/packages/shared/analytics/types.ts
+++ b/packages/shared/analytics/types.ts
@@ -63,4 +63,11 @@ export type TrackEvents = PositionTrackEvents & {
   ['unity_initializing_start']: { renderer_version: string }
   ['unity_initializing_end']: { renderer_version: string; loading_time: number }
   ['scene_start_event']: { scene_id: string; time_since_creation: number }
+
+  ['Friend request approved']: Record<string, unknown> // {}
+  ['Friend request rejected']: Record<string, unknown> // {}
+  ['Friend request cancelled']: Record<string, unknown> // {}
+  ['Friend request received']: Record<string, unknown> // {}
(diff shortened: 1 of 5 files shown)


packages/shared/analytics/types.ts
```diff
@@ -63,4 +64,11 @@ export type TrackEvents = PositionTrackEvents & {
   ['unity_initializing_start']: { renderer_version: string }
   ['unity_initializing_end']: { renderer_version: string; loading_time: number }
   ['scene_start_event']: { scene_id: string; time_since_creation: number }
+
```
gonpombo8, line 67: can we delete the ADD_FRIEND event that's deprecated and not in use ?

gonpombo8 (approved):

packages/shared/analytics/types.ts
```diff
@@ -63,4 +63,11 @@ export type TrackEvents = PositionTrackEvents & {
...
+  ['Friend request approved']: Record<string, unknown> // {}
+  ['Friend request rejected']: Record<string, unknown> // {}
+  ['Friend request cancelled']: Record<string, unknown> // {}
+  ['Friend request received']: Record<string, unknown> // {}
+  ['Friend request sent']: Record<string, unknown> // {}
+  ['Friend deleted']: Record<string, unknown> // {}
```
menduz, line 72: Can these be empty objects instead? Records are too generic and it defeats the purpose of having types

AjimenezDCL: Of course, I was just following the example with other empty events: `['motd_failed']: Record<string, unknown> // {}`

PR #272, opened 2022-04-18
fix: the parsing to disable and enable a feature from the url params

kuruk-mm:
# What? 

Fix https://github.com/decentraland/unity-renderer/issues/2171

We're using from the Launcher the URLParams to create the URL to add parameters like:

```
DISABLE_ASSET_BUNDLES
DISABLE_WEARABLES_ASSET_BUNDLES
```

And the output URL using URLParams is:

```
&DISABLE_ASSET_BUNDLES=&DISABLE_WEARABLES_ASSET_BUNDLES=
```

This is totally valid based on http://www.faqs.org/rfcs/rfc2396.html (thanks @leanmendoza)

But the kernel is ignoring it due to incorrect parsing.

### DIFF
--- a/packages/shared/meta/selectors.ts
+++ b/packages/shared/meta/selectors.ts
@@ -79,13 +79,13 @@ export function getFeatureFlags(store: RootMetaState): FeatureFlag {
     }
   }
   if (location.search.length !== 0) {
-    const flags = location.search.substr(1, location.search.length - 1).split('&')
-    flags.forEach((element) => {
-      if (element.includes(`DISABLE_`)) {
-        const featureName = element.replace('DISABLE_', '').toLowerCase()
+    const flags = new URLSearchParams(location.search)
+    flags.forEach((_, key) => {
+      if (key.includes(`DISABLE_`)) {
+        const featureName = key.replace('DISABLE_', '').toLowerCase()
         featureFlag.flags[featureName] = false
-      } else if (location.search.includes(`ENABLE_`)) {
-        const featureName = element.replace('ENABLE_', '').toLowerCase()
+      } else if (key.includes(`ENABLE_`)) {
+        const featureName = key.replace('ENABLE_', '').toLowerCase()
         featureFlag.flags[featureName] = true
       }
     })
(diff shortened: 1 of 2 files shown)


packages/shared/meta/selectors.ts
```diff
@@ -79,13 +79,13 @@ export function getFeatureFlags(store: RootMetaState): FeatureFlag {
...
-    const flags = location.search.substr(1, location.search.length - 1).split('&')
-    flags.forEach((element) => {
-      if (element.includes(`DISABLE_`)) {
-        const featureName = element.replace('DISABLE_', '').toLowerCase()
+    const flags = new URLSearchParams(location.search)
+    flags.forEach((_, key) => {
+      if (key.includes(`DISABLE_`)) {
+        const featureName = key.replace('DISABLE_', '').toLowerCase()
         featureFlag.flags[featureName] = false
-      } else if (location.search.includes(`ENABLE_`)) {
```
kuruk-mm, line 87: This could be causing other problems.
If you use `ENABLE_x` you were adding a feature to the feature map for every parameter that you had in the url.

So for example with the following URLs:
```
&ENABLE_x&hello&how&are&you
```
the featureFlag.flags dictionary would be:

```
flags['x'] = true
flags['hello'] = true
flags['how'] = true
flags['are'] = true
flags['you'] = true
```

packages/shared/meta/selectors.ts
```diff
@@ -79,13 +79,13 @@ export function getFeatureFlags(store: RootMetaState): FeatureFlag {
     }
   }
   if (location.search.length !== 0) {
-    const flags = location.search.substr(1, location.search.length - 1).split('&')
-    flags.forEach((element) => {
-      if (element.includes(`DISABLE_`)) {
-        const featureName = element.replace('DISABLE_', '').toLowerCase()
+    const flags = new URLSearchParams(location.search)
```
gonpombo8, line 82: can you add some unit tests to check that this works as expected ? Thanks :)

gonpombo8 (approved):

PR #279, opened 2022-04-26
fix: Set the WORLD_CHAT_WINDOW visible param as FALSE by default

sandrade-dcl:
# What? 
We have changed the WORLD_CHAT_WINDOW `visible` param as `false` by default.

# Why? 
When the application starts, kernel sends a message to Unity saying the chat should be visible by default, but the behaviour of the [**new redesign of the chat feature**](https://github.com/decentraland/unity-renderer/pull/1994) that we're currently implementing should be hidden by default and only be visible when the user clicks the chat icon or press enter.

# How to test it?
1. Go to https://play.decentraland.zone/?renderer-branch=feat/SocialBar&kernel-branch=fix/set-world-chat-window-visible-state-as-false-by-default
2. Check the new social window in the taskbar is initially closed.

### DIFF
--- a/packages/entryPoints/index.ts
+++ b/packages/entryPoints/index.ts
@@ -51,7 +51,7 @@ function configureTaskbarDependentHUD(i: IUnityInterface, voiceChatEnabled: bool
       enableQuestPanel: isFeatureEnabled(store.getState(), FeatureFlags.QUESTS, false)
     }
   )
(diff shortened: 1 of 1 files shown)


davidejensen (approved):

menduz (approved):

menduz:
@sandrade-dcl feel free to merge when needed

gonpombo8 (approved):

sandrade-dcl:
This PR is no longer needed as we already have the same change included in https://github.com/decentraland/kernel/pull/282

PR #282, opened 2022-04-29
feat: Retry the friends initialization

sandrade-dcl:
# What? 
Now the initialization of the friends feature is being retried when it fails. That retry is carried out using an exponential time increment up to a maximum allowed.


menduz:
To prevent unnecessary messages and moving the control to the renderer, I'd implement this as:

```typescript
while(true) {
try {
  yield call(initialize)
} catch {
  if (attempt++ < MAX_ATTEMPTS) {
    yield delay(N)
  } else {
    break
  }
}
```

packages/shared/friends/sagas.ts
```diff
@@ -75,13 +78,25 @@ function* initializeFriendsSaga() {
...
-      const synapseUrl: string = yield select(getSynapseUrl)
-      yield call(initializePrivateMessaging, synapseUrl, identity)
-    } catch (e) {
-      logger.error(`error initializing private messaging`, e)
+    let secondsToRetry = MIN_TIME_BETWEEN_FRIENDS_INITIALIZATION_RETRIES_MILLIS
+
+    while(true) {
+      try {
+        const synapseUrl: string = yield select(getSynapseUrl)
+        yield call(initializePrivateMessaging, synapseUrl, identity)
```
menduz, line 86: when does this loop ends?

menduz: but it doesn't stop if it worked, or I am missing it

menduz: Here are two examples of controlled infinite loops https://github.com/decentraland/kernel/blob/chore/new-comms-sagas/packages/shared/comms/sagas.ts#L114-L154

menduz: I'd add something like 
```ts
const client: SocialAPI | null = yield select(getClient)
const isLoggedIn: boolean = client && (yield apply(client, client.isLoggedIn) || false
if (isLoggedIn) {
  // reset timers
  // yield delay(initialDelay)
} else {
  yield call(initializePrivateMessaging, synapseUrl, identity)
}
```

packages/shared/friends/sagas.ts
```diff
@@ -48,6 +48,7 @@ import { ensureFriendProfile } from './ensureFriendProfile'
 import { getSynapseUrl } from 'shared/meta/selectors'
 import { store } from 'shared/store/isolatedStore'
 import { trackEvent } from '../analytics'
+import { apply } from 'redux-saga-test-plan/matchers'
```
menduz, line 51:
```suggestion
import { apply } from 'redux-saga/effects'
```

sandrade-dcl: Ups, fixed!

menduz (dismissed):

packages/shared/friends/sagas.ts
```diff
@@ -26,158 +24,166 @@ import {
...
+    yield call(waitForRealmInitialized)
+    yield call(waitForRendererInstance)
+
+    const currentIdentity: ExplorerIdentity | undefined = yield select(getCurrentIdentity)
+    const client: SocialAPI | null = yield select(getSocialClient)
+    const isLoggedIn: boolean = (currentIdentity && client && (yield apply(client, client.isLoggedIn, []))) || false
+
+    const shouldRetry = !isLoggedIn
+
+    if (shouldRetry) {
```
moliva, line 106: Should we have a limited number of retries instead of doing it indefinitely?

menduz: The idea was to have it indefinitely recovery. but the client has a problem that prevents it from happening if some events are pending to be sent to matrix. I'd recommend to stop using the dcl-social-api and using matrix-js-sdk from kernel directly instead

PR #285, opened 2022-05-02
feat: CRDT message to renderer

pbosio:
# What? 
...

# Why? 
...


leanmendoza:
The approach looks good to me! I suggest the previous comment as optional and another question:
can we change the `crdt` for `wire` o `binary`. CRDT would be the first message but maybe we'll add another type of messages (like raycasts probably?)


_nativeMsgBridge.**crdt**Message(sceneId, message)_ => nativeMsgBridge.**wire**Message(sceneId, message)

leanmendoza (approved):

PR #297, opened 2022-05-10
hotfix: also hit lambdas to check liveness

menduz:
# What? 
...

# Why? 
...


packages/shared/dao/index.ts
```diff
@@ -51,6 +51,10 @@ export function isPeerHealthy(peerStatus: Record<string, HealthStatus>) {
   )
 }
 
+export function lambdasStatusUrl(domain: string) {
+  return `${domain}/lambdas/status`
```
agusaldasoro, line 55: please use and check that the 3 services are "Healthy"
```suggestion
  return `${domain}/lambdas/health`
```

pedrotambo (dismissed):
LGTM

PR #300, opened 2022-05-12
feat: Make kernel capable of loading single items from the builder server

LautaroPetaccio:
# What? 
This PR adds a method and a new query string to load a set of items from the builder server by id.

# Why? 
We need a way to load sets of items into the world so curators and creators can view small sets of items. This feature is intended to be used with third party collection that contain thousands of items and can't be loaded as collections without pagination.

### DIFF
--- a/packages/shared/catalogs/sagas.ts
+++ b/packages/shared/catalogs/sagas.ts
@@ -8,7 +8,8 @@ import {
   DEBUG,
   ETHEREUM_NETWORK,
   BUILDER_SERVER_URL,
-  rootURLPreviewMode
+  rootURLPreviewMode,
+  WITH_FIXED_ITEMS
 } from 'config'
 
 import defaultLogger from 'shared/logger'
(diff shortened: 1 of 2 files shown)


packages/shared/catalogs/sagas.ts
```diff
@@ -184,6 +197,26 @@ async function fetchWearablesByFilters(filters: WearablesRequestFilters, client:
   return client.fetchWearables(filters)
 }
 
+/**
+ * Fetches a single item from the Builder Server.
+ */
+async function fetchWearableByIdFromBuilder(uuidsItems: string[], identity: ExplorerIdentity): Promise<WearableV2[]> {
```
nachomazzara, line 203: Maybe we need to name `fetchWearableByIdFromBuilder` plural `fetchWearablesByIdFromBuilder`?

LautaroPetaccio: Done!

menduz (approved):

PR #306, opened 2022-05-23
fix: hot-reload in smart-wearables

leanmendoza:
# What? 
This only brings back the `await sleep` set between the destruction and creation of Portable Experiences coded in:
https://github.com/decentraland/kernel/blob/039d162fdfd587106ab86c4018997cb2498d061f/packages/unity-interface/dcl.ts#L214

# Why? 
The `spawnPx..` right now is ignored if it's called immediately after its destruction.

## NOTE: this is not a solution, just a workaround until the real problem be taken.

### DIFF
--- a/packages/unity-interface/portableExperiencesUtils.ts
+++ b/packages/unity-interface/portableExperiencesUtils.ts
@@ -14,6 +14,7 @@ import { getUnityInstance } from './IUnityInterface'
 import { resolveUrlFromUrn } from '@dcl/urn-resolver'
 import { store } from 'shared/store/isolatedStore'
 import { addScenePortableExperience, removeScenePortableExperience } from 'shared/portableExperiences/actions'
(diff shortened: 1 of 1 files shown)


packages/unity-interface/portableExperiencesUtils.ts
```diff
@@ -127,6 +128,10 @@ export async function declareWantedPortableExperiences(pxs: StorePortableExperie
     }
   }
 
+  // TODO: this is an ugh workaround, fix controlling the scene lifecycle
+  // knowing when the scene was completly removed and then re-spawn it
+  await sleep(250)
```
gonpombo8, line 133: this sounds super arbitrary.
Can we attach the dispose to a promise so we can wait to all the workers to be removed ?  Is it possibe ?
```ts 
await forceStopSceneWorker(scene.worker))

```

PR #313, opened 2022-05-31
fix: adding error handling and retrying to private chat message retri…

### LINKED ISSUE #315: Fix issue with account_data updates provoking CORS error screen
This is currently happening when starting up the client and many concurrent calls to the account_data PUT endpoint are made. The main flow keeps running towards friend messages fetching when this ends up erring.
Image

moliva:
Fixes #315.

# What? 
...

# Why? 
...


packages/shared/friends/sagas.ts
```diff
@@ -239,25 +240,60 @@ function* configureMatrixClient(action: SetMatrixClient) {
   yield Promise.all(
```
menduz, line 240: Maybe wrap this in a try catch to prevent bubbling up the exception to a fatal error
```suggestion
  yield Promise.all(
```

menduz (approved):

PR #334, opened 2022-06-09
PR review suggestions

menduz:
- Simplified module loading in Scene runtime
- Using new RPC in the avatars' scene
- Use makefile features to auto-detect .proto files
- .proto files moved to the same folder as the generated files
- Fixed casing in .proto descriptions


packages/shared/apis/host/SocialController.ts
```diff
@@ -1,21 +1,28 @@
...
+          }
+        })
+        const observer = avatarMessageObservable.add((event) => {
+          messageQueue.enqueue({
+            event: event.type,
+            payload: JSON.stringify(event)
+          })
+        })
+        return messageQueue
+      }
```
menduz, line 25: @leanmendoza

PR #349, opened 2022-06-18
feat: remove in-world builder and simplify data flows

menduz:
REQUIRES: https://github.com/decentraland/protocol/pull/6

This PR removes in-world builder support and simplifies the land loading mechanisms.

The result of it, is that now either portable experiences and scenes share the same descriptor data structure `LoadableScene`. It is intended to be the only required thing to load any scene and it looks like this:

```ts
import { Entity, Scene } from '@dcl/schemas'

// From the entity, only the following are used:
type Entity = {
  contents: ContentMapping
  metadata: any // in practice, it is of type Scene
}

type LoadableScene = {
  id: string // URN of the deployed entity
  baseUrl: string // baseUrl to look for the content files
  entity: Entity
  parentCid: string // "spawner" scene, the same concept as parent process in OS. (this name will change)
}
```

This PR also uses the new active entities endpoint, which is two orders of magnitude faster. It should also help reduce the DB load of the content servers. While changing the logic for that refactor, the scene lifecycle manager got also simplified. Now the process of loading scenes and empty parcels got _extremely_ simplified removing coupling between classes and unnecessary dependency injections.

Also the many abstractions on top of the scenes (UnityScene, UnityParcelScene, UnityPortableExperienceScene) were simplified into only one class.

# TODO

- [x] check what does the renderer do with LoadParcelScenes now that the type has changed
- [x] Verify that absolute(world) camera position still works as expected


packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -25,95 +22,75 @@ export async function startSceneRuntime(client: RpcClient) {
...
+    await EngineAPI.sendBatch({ actions: [initMessagesFinished()] })
+    throw new Error(`SDK: Error while loading scene. Main file missing.`)
+  }
+
+  const url = resolveMapping(mapping.hash, mappingName, bootstrapData.baseUrl)
   const codeRequest = await fetch(url)
 
   if (!codeRequest.ok) {
-    EngineAPI.sendBatch({ actions: [initMessagesFinished()] }).catch((err) => devToolsAdapter.error(err))
+    await EngineAPI.sendBatch({ actions: [initMessagesFinished()] })
```
gonpombo8, line 61: try/catch ?

packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -122,88 +99,89 @@ export async function startSceneRuntime(client: RpcClient) {
...
-   * Handle the pull events response calling the eventReceiver
-   */
-  function processEvents(req: PullEventsResponse) {
-    for (const e of req.events) {
-      eventReceiver(EventDataToRuntimeEvent(e))
+    for (const cb of onEventFunctions) {
+      try {
+        cb(event)
+      } catch (err) {
+        console.error(err, { event })
```
gonpombo8, line 112:
```suggestion
        defaultLogger.error(err, { event })
```

packages/shared/comms/peers.ts
```diff
@@ -1,4 +1,4 @@
-import { Observable } from '@dcl/legacy-ecs'
+import { Observable } from 'mz-observable'
```
gonpombo8, line 1: 👀

leanmendoza (approved):
LGTM! I pay special attention to the SceneRuntime part and I tested the use cases and everything is OK!

packages/scene-system/sdk/SceneRuntime.ts
menduz: Next line is a throw

kuruk-mm (dismissed):
LGTM!

kuruk-mm (approved):

PR #354, opened 2022-06-22
fix: null camera target set as undefined

leanmendoza:
Also set .org rollout to 100 (chore)

# What? 
...

# Why? 
Fix the condition the cameraTarget is sent as null. Before the usage of @dcl/rpc this was working because the field wired as is and the remote procedure received the `null`, so the ifs were working fine. Now the field is serialized and the fields have to be able to be serialized, or if it's not going to wire, be undefined.

### DIFF
--- a/.github/workflows/set-rollouts.yml
+++ b/.github/workflows/set-rollouts.yml
@@ -58,4 +58,4 @@ jobs:
           # Rollout information
           deploymentDomain: "play.decentraland.org"
           deploymentName: "@dcl/kernel"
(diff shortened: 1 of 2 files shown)


gonpombo8 (approved):

pravusjif:
Tested at https://play.decentraland.org/?kernel-branch=fix%2Fcamera-target&NETWORK=ropsten&position=70,-63&LOS=2 and I confirm the teleport bug is now fixed, great job!

PR #356, opened 2022-06-24
feat: chat metrics back in kernel

AjimenezDCL:
Chat metrics are back in kernel


packages/shared/chat/sagas.ts
```diff
@@ -71,9 +72,12 @@ function* trackEvents(action: PayloadAction<MessageEvent, ChatMessage>) {
   const { type, payload } = action
   switch (type) {
     case SEND_MESSAGE: {
-      trackEvent('Control Send chat message', {
+      trackEvent('Send chat message', {
         messageId: payload.messageId,
+        from: hasWallet() ? ChatMessagePlayerType.WALLET : ChatMessagePlayerType.GUEST,
+        to: ChatMessagePlayerType.WALLET,
```
moliva, line 78: Remember this could also be a public chat (currently `#nearby`) sent through comms, please check that the `payload.messageType === Private` for this to be WALLET, else it might be better to set this to `undefined` or another enum value like `Nearby`.

AjimenezDCL: Done!

packages/shared/chat/sagas.ts
```diff
@@ -75,9 +75,9 @@ function* trackEvents(action: PayloadAction<MessageEvent, ChatMessage>) {
       trackEvent('Send chat message', {
         messageId: payload.messageId,
         from: hasWallet() ? ChatMessagePlayerType.WALLET : ChatMessagePlayerType.GUEST,
-        to: ChatMessagePlayerType.WALLET,
+        to: payload.messageType === ChatMessageType.PRIVATE ? undefined : ChatMessagePlayerType.WALLET,
```
menduz, line 78: This logic is not clear, what does "to" stands for? From my understanding if the message type is PRIVATE, then `to` should be WALLET, what is the second option? what does undefined stands for?

AjimenezDCL: "to" is the recipient of the message. I agree `recipient` would be more describing but since we already have events tracked with that field as "to" I would suggest to keep it this way.

You are right! I misplaced the ternary operator, it should be WALLET for PRIVATE msgs and undefined (because there's no recipient) for the rest.

menduz (approved):

PR #357, opened 2022-06-27
fix: slow changerealm with avatar batch events

leanmendoza:
# What? 
...

# Why? 
...


gonpombo8 (approved):

menduz (approved):
Looks good, if it works 🚢  it

PR #363, opened 2022-07-04
feat: add friends lazy load

jmoguilevsky:
# What? 
This PR implements lazy loading of friends following the API defined in [this document](https://www.notion.so/decentraland/RFC-Lazy-loading-for-social-data-06ff22ce0545464b83da3e4b0457e901)


packages/shared/friends/sagas.ts
```diff
@@ -378,6 +312,42 @@ function* refreshFriends() {
...
+      unseenPrivateMessages
+    }
+
+    console.log('____ initMessage ____', initMessage)
+
+    getUnityInstance().InitializeFriends(initMessage)
+
+    const userIds = Object.values(socialInfo).map((info) => info.userId)
+
+    yield ensureFriendProfile(userIds).catch(logger.error)
```
moliva, line 349: I think this should be done before initializing friends in renderer.

packages/shared/profiles/ProfileAsPromise.ts
```diff
@@ -11,53 +11,75 @@ const PROFILE_SOFT_TIMEOUT_MS = 5000
 // We reject the profile promise if more time than this has passed
 const PROFILE_HARD_TIMEOUT_MS = COMMS_PROFILE_TIMEOUT + 20000
 
-export function ProfileAsPromise(userId: string, version?: number, profileType?: ProfileType): Promise<Avatar> {
+export function ProfilesAsPromise(userIds: string[], version?: number, profileType?: ProfileType): Promise<Avatar[]> {
```
menduz, line 14: For clarity purposes and to prevent accidents with accesors, I'd recommend to have two versions of this function. The new `ProfilesAsPromise(): Array<Profile>` and the original `ProfileAsPromise(): Profile`. That will simplify a lot the changes required to test this PR

packages/shared/profiles/reducer.ts
```diff
@@ -41,6 +46,51 @@ export function profileReducer(state?: ProfileState, action?: AnyAction): Profil
...
+      }
+
+    case PROFILES_FAILURE:
+      const { userIds } = (action as ProfilesFailureAction).payload
+      const newProfilesState = userIds.reduce((newState, userId) => {
+        return {
+          ...newState,
+          [userId]: { status: 'error', data: action.payload.error }
+        }
+      }, {})
```
menduz, line 85: Kernel code prefers vanilla loops when possible, mainly for performance and secondly for readability. Please apply to all occurrences of .reduce

```suggestion
      const newProfilesState: Record<string, ProfileUserInfo> = {}
      for (const userId of userIds) {
        newProfilesState[userId] = { status: 'error', data: action.payload.error }
      }
```

packages/shared/profiles/sagas.ts
```diff
@@ -78,11 +84,19 @@ export function* profileSaga(): any {
...
   yield fork(handleCommsProfile)
   yield debounce(200, DEPLOY_PROFILE_REQUEST, handleDeployProfile)
   yield takeEvery(SAVE_PROFILE, handleSaveLocalAvatar)
 }
 
+function* forwardProfilesToRenderer(action: ProfilesSuccessAction) {
+  for (const profile of action.payload.profiles) {
+    yield put(sendProfileToRenderer(profile.userId))
+  }
+}
```
menduz, line 98: This should be centralized in one single place, I'd prevent duplicating this action if it was not moved from its original location

packages/shared/profiles/sagas.ts
```diff
@@ -183,52 +197,91 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
-    let avatar = profiles.avatars[0]
+  const identity: ExplorerIdentity | undefined = yield select(getCurrentIdentity)
 
-    if (avatar) {
-      avatar = ensureAvatarCompatibilityFormat(avatar)
-      if (!validateAvatar(avatar)) {
-        defaultLogger.warn(`Remote avatar for user is invalid.`, userId, avatar, validateAvatar.errors)
-        return null
-      }
+  if (!identity) throw new Error("Can't fetch profile if there is no ExplorerIdentity")
```
menduz, line 205: Why?

packages/shared/profiles/sagas.ts
```diff
@@ -183,52 +197,91 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
 
-export async function profileServerRequest(userId: string, version?: number) {
+export async function profilesServerRequest(userIds: string[], version?: number): Promise<RemoteProfile[]> {
   const state = store.getState()
   const catalystUrl = getCatalystServer(state)
 
   try {
-    let url = `${catalystUrl}/lambdas/profiles?id=${userId}`
+    let url = `${catalystUrl}/lambdas/profiles`
     if (version) url = url + `&version=${version}`
```
menduz, line 265: Version is only used on a per-profile basis, should not be present when fetching many

packages/shared/profiles/sagas.ts
```diff
@@ -183,52 +197,91 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
 
-    const response = await fetch(url)
+    const response = await fetch(url, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({ ids: userIds })
```
menduz, line 272: Version is _very_ important to kernel and renderer. Please account for that, it is a caching mechanism

packages/shared/profiles/sagas.ts
```diff
@@ -183,52 +197,91 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
+    })
 
     if (!response.ok) {
       throw new Error(`Invalid response from ${url}`)
     }
 
     const res = await response.json()
 
-    return res[0] || { avatars: [] }
+    return res || [{ avatars: [], timestamp: Date.now }]
```
menduz, line 281: Shouldn't the multiple result be inside `avatars`? why do we have an array inside another array?

packages/shared/profiles/types.ts
```diff
@@ -24,3 +24,8 @@ export enum ProfileType {
   LOCAL = 0,
   DEPLOYED = 1
 }
+
+export interface RemoteProfile {
```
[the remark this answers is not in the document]
menduz: ➕

packages/shared/profiles/sagas.ts
```diff
@@ -156,7 +170,7 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
       // first fetch avatar through comms
       (shouldFetchViaComms && (yield call(requestProfileToPeers, commsContext, userId, version))) ||
       // and then via catalyst
-      (shouldLoadFromCatalyst && (yield call(getRemoteProfile, userId, version))) ||
+      (shouldLoadFromCatalyst && (yield call(getRemoteProfiles, [userId], version)))[0] ||
```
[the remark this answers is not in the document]
jmoguilevsky: You do, it's being assigned to the variable `avatar`

packages/shared/profiles/sagas.ts
jmoguilevsky: That's because the response from the Backend responds a profile that contains an array of avatars, as seen in the new type `RemoteProfile` it's always been that way

jmoguilevsky: A profile can contain multiple avatars

packages/shared/profiles/sagas.ts
jmoguilevsky: Based on multiple of these comments I think the best we can do is to have just the two mechanisms separated, on the one hand, have just a single out fetch profile that will account for the version (which is the one that already existed), and the new one which will only be for initialization of friendships which will be able to fetch multiple profiles at the same time, does that sound right?

packages/shared/profiles/reducer.ts
jmoguilevsky: Okay, will replace the reduces

packages/shared/profiles/sagas.ts
jmoguilevsky: We just kept the same check that was being performed in the original method to prevent any errors
https://github.com/decentraland/kernel/blob/79edd230d0c2b5cc43b7a6d7a1bcc9e4050b7d3a/packages/shared/profiles/sagas.ts#L147

packages/shared/profiles/sagas.ts
jmoguilevsky: This is not being used, it was copied and pasted for an explanation during a call, it's going to be erased

packages/shared/profiles/sagas.ts
menduz: sounds perfect, thanks. I assume that once the new mechanism is working we could implement the fetch simple as a wrapper for the fetch multiple

packages/shared/profiles/sagas.ts
menduz: ok, thats a new one. do you know how does it work? because you can only deploy one profile per address

packages/shared/profiles/types.ts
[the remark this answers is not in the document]
jmoguilevsky: But the profile is only one, it just has many avatars

packages/shared/profiles/sagas.ts
jmoguilevsky: I don't really know why the response is built that way, I'll sync with the explorer's and platform team to understand why the decision of multiple avatars per profile.

On another hand what I do know is that currently only the first avatar of each profile is being used as seen here https://github.com/decentraland/kernel/blob/79edd230d0c2b5cc43b7a6d7a1bcc9e4050b7d3a/packages/shared/profiles/sagas.ts#L438

packages/shared/dao/sagas.ts
```diff
@@ -167,7 +167,8 @@ function* selectRealm() {
 
   if (!realm) debugger
 
-  return realm
+  // return realm
+  return { hostname: 'https://peer-testing.decentraland.org', protocol: 'v2', serverName: 'testing' }
```
menduz, line 171: `?realm=v2~peer-testing.decentraland.org` in the URL does the trick

jmoguilevsky: I know, but for some reason, it was still changing it back

packages/shared/profiles/sagas.ts
jmoguilevsky: So after syncing with the platform team, apparently they just have an array BUT it's always a size 1 array, that's why everywhere you'll find `profile.avatars[0]` because there's no situation where avatars can have more than one. 
Still, changing the API would be a breaking change, and that's why it wasn't done (there are too many dependencies on that endpoint now for it to change)

packages/shared/friends/sagas.ts
```diff
@@ -388,22 +403,218 @@ function* refreshFriends() {
...
+    updateTotalUnseenMessagesByUserPayload.unseenPrivateMessages.push({
+      count: conversation.conversation.unreadMessages?.length || 0,
+      userId: conversation.conversation.userIds![1]
+    })
+  }
+
+  getUnityInstance().UpdateTotalUnseenMessagesByUser(updateTotalUnseenMessagesByUserPayload)
+}
+
+export function getFriendsWithDirectMessages(request: GetFriendsWithDirectMessagesPayload) {
```
Julieta11, line 565: - We have some console logs here, not sure if we want to keep them. 
- Also, I think we're not lazy loading the list of conversations. I mean we should be making use of request.limit and request.skip, shouldn't we?

moliva: Let's go ahead and remove all `console#log` as well as any commented out code.

packages/unity-interface/BrowserInterface.ts
```diff
@@ -87,16 +87,11 @@ import {
   getPrivateMessages,
   markAsSeenPrivateChatMessages
 } from 'shared/friends/sagas'
+import { getMatrixIdFromUser } from 'shared/friends/utils'
 
 declare const globalThis: { gifProcessor?: GIFProcessor }
 export const futures: Record<string, IFuture<any>> = {}
 
-// ** TODO - move to friends related file - moliva - 15/07/2020
```
moliva, line 81: Thanks! 🙌🏻

packages/shared/friends/sagas.ts
```diff
@@ -374,9 +360,40 @@ function* refreshFriends() {
...
+
+    const initFriendsMessage: FriendsInitializationMessage = {
+      totalReceivedRequests: requestedFromIds.length
+    }
+    const initChatMessage: FriendsInitializeChatPayload = {
+      totalUnseenMessages
+    }
+
+    defaultLogger.log('____ initMessage ____', initFriendsMessage)
+    defaultLogger.log('____ initChatMessage ____', initChatMessage)
```
moliva, line 387: Should we use change these to `debug` or `info` maybe?

moliva:
Closing in favor of #484.

PR #373, opened 2022-07-11
fix: legacy transform serialization

pbosio:
add code to branch `Transform` serialization:
-current: protobuf -> base64
-new: arraybuffer -> base64
renderer will be sending `useBinaryTransform` flag with it first message to be able to select which serialization method to use and keep backward compatibility with previous versions of the renderer.
we are adding this because updating `protoc` in the renderer is causing a performance regression due to the protobuff serialization


kuruk-mm (approved):
Patobinary

packages/scene-system/sdk/Utils.ts
```diff
@@ -60,6 +69,39 @@ function serializeTransform(transform: Transform): string {
...
+  // If we don't cap these vectors, scenes may trigger a physics breakdown when messaging enormous values
+  const cappedVector = new Vector3(
+    Math.fround(transform.position.x),
+    Math.fround(transform.position.y),
+    Math.fround(transform.position.z)
+  )
+  capVector(cappedVector, VECTOR3_MEMBER_CAP)
+
+  let offset: number = 0
+  transformView.setFloat32(offset, cappedVector.x, true)
```
leanmendoza, line 83: thanks for specifying the endianness :pray:

packages/unity-interface/BrowserInterface.ts
```diff
@@ -263,9 +264,12 @@ export class BrowserInterface {
     trackEvent('performance report', perfReport)
   }
 
-  public SystemInfoReport(data: SystemInfoPayload) {
+  // TODO: remove useBinaryTransform after ECS7 is fully in prod
```
leanmendoza, line 267: Could we track this with an issue in the ECS7 milestone?

kuruk-mm (approved):
Great!

PR #379, opened 2022-07-19
Feat: add requests lazy loading

### LINKED ISSUE #343: Load friend requests list on demand
- Renderer->kernel message to load friend requests (to and from current user) (paginated)
- Kernel->renderer message with friend requests (to and from current user) (paginated)
- Implementation in friends saga

jmoguilevsky:
Closes #343 

# What? 
Add friend requests lazy loading

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -35,7 +35,7 @@ import { lastPlayerPosition } from 'shared/world/positionThings'
 import { waitForRendererInstance } from 'shared/renderer/sagas-helper'
 import { getProfile, getProfilesFromStore } from 'shared/profiles/selectors'
 import { ExplorerIdentity } from 'shared/session/types'
-import { SocialData, FriendsState } from 'shared/friends/types'
+import { SocialData, FriendsState, FriendRequest } from 'shared/friends/types'
 import {
   getSocialClient,
   findPrivateMessagingFriendsByUserId,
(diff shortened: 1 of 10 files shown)


test/unit/friends.saga.test.ts
```diff
@@ -124,4 +144,40 @@ describe('Friends sagas', () => {
...
+    afterEach(() => {
+      sinon.restore()
+      sinon.reset()
+    })
+
+    describe("When there're sent and received friend requests", () => {
+      it("Should call unity with the declared parameters", () => {
+        const request: GetFriendRequestsPayload = {
+          sentLimit: 10,
+          sentSkip: 0,
```
jmoguilevsky, line 165: I'd do a test that has a skip and a limit to show that the slice is bringing the expected information as done in the getFriends tests, otherwise, great job!

packages/shared/friends/sagas.ts
```diff
@@ -402,6 +416,28 @@ export function getFriends(request: GetFriendsPayload) {
...
+  const toFriendRequests = friends.toFriendRequests.slice(
+    request.sentSkip,
+    request.sentSkip + request.sentLimit
+  )
+
+  const addFriendRequestsPayload: AddFriendRequestsPayload = {
+    requestedTo: toFriendRequests.map((friend) => friend.userId),
+    requestedFrom: fromFriendRequests.map((friend) => friend.userId),
+    totalReceivedFriendRequests: fromFriendRequests.map((friend) => friend.userId).length,
+    totalSentFriendRequests: toFriendRequests.map((friend) => friend.userId).length
```
moliva, line 435: It's not necessary to do the mapping of both arrays if we are only taking the length (also, both of them were just mapped two lines above).

packages/shared/types.ts
```diff
@@ -538,6 +538,13 @@ export interface GetFriendsPayload {
   skip: number // amount of entries to skip
 }
 
+export interface GetFriendRequestsPayload {
+  sentLimit: number // max amount of entries to receive
+  sentSkip: number // the amount of sent requests to skip
+  receivedLimit: number // max amount of entries to receive
```
moliva, line 544: What's the difference with the `sentLimit`, let's clarify that for the four fields.

jmoguilevsky: received are the requests that you have received, only the description is wrong

moliva (approved):

moliva:
In the following tasks, I would like to start thinking how to break up and extract functions away from our current sagas to dedup some of our code and have an easier to follow flow.

Issue #339, opened 2022-06-13
Voice Chat system is not working correctly


sandrade-dcl:
## Description
After doing several tests with several people, we have reached the conclusion that something in Kernel side broke the Voice Chat in some point and it is not related to .zone or .org as we though at the beginning. We have found some cases (like @olavra) where this Voice Chat issue is happening ALWAYS:

1. He try to activate the microphone clicking the microphone button.
2. The browser ask about the needed mic permissions and he allows them.
3. From this point on, the microphone button keeps indefinitely pressed  but he never receives any update from Kernel. All the voice chat system seems to be totally deactivated and he can't send audio neither receive it from anyone.

![image.png](https://images.zenhubusercontent.com/5eb2b8ca65c88f6e391ef16e/1f65a711-6663-46c9-8d24-3decd0dd0abc)

However for other users (like @sandrade-dcl) the Voice Chat is working perfectly... it is very weird.

We have also checked it doesn't mind if the test is done from Mac or from Windows, if it is .org or .zone... we have gotten different results depending the user.

## Acceptance criteria
- [ ] The Voice Chat system should work fine for all users.

Suduck:
For me, it works from the UI perspective, however, people from around me can't hear me

PR #380, opened 2022-07-20
fix: voice-chat

kuruk-mm:
Fix https://github.com/decentraland/sdk/issues/232
Fix https://github.com/decentraland/kernel/issues/339

# What?
After the changes of the comms sagas (https://github.com/decentraland/kernel/pull/261). The voice chat was being blocked by the `Autoplay Policy in Chrome` (https://developer.chrome.com/blog/autoplay/#webaudio).

For this reason, the `process` function from `AudioWorkletProcessor` was not being executed.
The kernel was sending to the Renderer just a `recording=true` notification. And the `recording=false` was being executed from the `process` function from the `AudioWorkletProcessor`.
Causing the following issues:
https://github.com/decentraland/sdk/issues/232
https://github.com/decentraland/kernel/issues/339

To mitigate the problem. This PR adds a `timeout` system if the `AudioWorkletProcessor` doesn't respond to avoid the Renderer being stuck for any reason.

After that, to solve the real problem. We just changed the comms location to initialize after comms and after the World Context is up.


packages/shared/comms/voice-over-comms.ts
```diff
@@ -13,6 +13,8 @@ import { put } from 'redux-saga/effects'
 import { setupPeer } from './peers'
 import { shouldPlayVoice } from './voice-selectors'
 import { positionObservable, PositionReport } from 'shared/world/positionThings'
+import { getUnityInstance } from 'unity-interface/IUnityInterface'
```
kuruk-mm, line 16: I hesitate to put the `getUnityInstance` here in the comms package. @gonpombo8 what do you think?

gonpombo8 (approved):

PR #382, opened 2022-07-25
update to goerli

gonpombo8:
https://github.com/decentraland/urn-resolver/blob/main/src/contracts.ts#L156-L160
Need to update urn-resolver with the changes for goerli

# What? 
...

# Why? 
...


packages/shared/dao/reducer.ts
```diff
@@ -64,7 +64,7 @@ function realmProperties(realm: Realm, network: ETHEREUM_NETWORK | null): Partia
   // this condition exists until the BFF is finished and the "services" become interfaces
   if (realm.protocol === 'v1' || realm.protocol === 'v2') {
     return lighthouseBasedPartial(domain)
-  } else if (network === ETHEREUM_NETWORK.ROPSTEN) {
+  } else if (network === ETHEREUM_NETWORK.GOERLI) {
     return lighthouseBasedPartial('https://peer.decentraland.zone')
```
marianogoldman, line 68: We currently kept both networks: ropsten and goerli. `peer.decentraland.zone` and `peer-ue-2.decentraland.zone` stay on ropsten, and `peer-ap1.decentraland.zone` is on a separate DAO on goerli.
We did this so we can keep our current working environment while the new one gets to working state.

PR #384, opened 2022-07-25
feat: use /about and /stats/parcels to get the status of a catalyst

hugoArregui:
The current implementation has the following behavior:

- If the catalyst is blacklisted is not selected
- if the `realm` is provided but it's blacklisted, kernel chooses another one
- if the `CATALYST` param is provided and it's blacklisted, kernel connects anyway.


packages/shared/dao/index.ts
```diff
@@ -42,64 +42,89 @@ export async function fetchCatalystRealms(nodesEndpoint: string | undefined): Pr
...
-      return peerStatus[server] !== HealthStatus.HEALTHY
-    })
-  )
-}
-
-export function lambdasHealthUrl(domain: string) {
-  return `${domain}/lambdas/health`
-}
+async function fetchCatalystStatus(domain: string, denylistedCatalysts: string[]): Promise<Candidate | undefined> {
+  if (denylistedCatalysts.includes(domain)) return undefined
```
menduz, line 46: I think this line should be part of the `checkValidRealm` function instead

packages/shared/dao/index.ts
```diff
@@ -42,64 +42,92 @@ export async function fetchCatalystRealms(nodesEndpoint: string | undefined): Pr
...
-  let url = `${domain}/comms/status`
-  const queryParameters: string[] = []
+      return {
+        protocol: comms.protocol,
+        catalystName: configurations.realmName,
+        domain: domain,
+        status: aboutResponse.status,
+        elapsed: aboutResponse.elapsed!,
+        usersCount: bff.userCount ?? 0,
+        maxUsers: -1, // TODO
```
kuruk-mm, line 87: Can we refer this to an issue or have more context in the comment?

hugoArregui: Sorry, I'll remove the comment, we are not going to use maxUsers anymore, since now we have the feature flag for disabled catalyst

kuruk-mm (approved):
I didn't test it, but the code looks great!

kuruk-mm (approved):

PR #385, opened 2022-07-26
feat: allow to run global playground scene

leanmendoza:
This is an almost free approach to making a playground with a preview server.

# What? 
- Add CATALYST select in preview mode (it already exists for non-preview)
- Fix the scene.json fetch by adding the base URL (fix this use case)

# Why? 
Running a preview scene (with or without comms), it's possible to refresh the scene with a postMessage.
This enables a playground with the next approach:
- Run the preview as an `iframe` component with the same origin (this is important) https://github.com/decentraland/sdk-playground/tree/preview
- Because of the same origin policy, it's possible to get the context to postMessage from the parent of the iframe.
- Edit code in the editor, compiles and paste it into a bundled ecs7

Again, rendering the explorer-client in the react-app is possible thanks to the same origin and to adding the custom `html` preview.html (https://github.com/decentraland/sdk-playground/blob/preview/public/preview/index.html#L639)

There are a lot of code to clean, but this is only a demonstration :
https://sdk-playground.vercel.app/
image

In the back-end there is a preview server
https://ecs7-template.herokuapp.com


packages/entryPoints/index.ts
```diff
@@ -306,4 +311,12 @@ export async function startPreview(unityInterface: IUnityInterface) {
       handleServerMessage(message)
     }
   })
+
+  window.addEventListener('message', async (msg) => {
+    if (typeof msg.data === 'string') {
+      if (msg.data.startsWith('{')) {
```
gonpombo8, line 317: whats this magic ? We are sending the code here and also via the gloabLThis.Playground ?
Maybe this could be something more specific ? `msg.data === 'reload-playground'` ?

leanmendoza: A residual behviour, I'll modify it (it had the CLI messages behaviour replicated)

menduz (approved):

PR #386, opened 2022-07-28
Add conversations list

### LINKED ISSUE #345: Load friends with existing chats on demand
- ⚠ Review the feasibility of this without actually having to retrieve all messages for all friends
- Renderer->kernel message to load friends with existing conversations (at least one message) (paginated)
- Kernel->renderer message with friends with existing conversations (paginated)
- Implementation in friends saga

jmoguilevsky:
Add lazy loading of all existent conversations

Closes #345

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -427,7 +427,7 @@ export function getFriendRequests(request: GetFriendRequestsPayload) {
   getUnityInstance().AddFriendRequests(addFriendRequestsPayload)
 }
 
-export async function getUnseenMessagesByUser() {
+export function getUnseenMessagesByUser() {
   const client: SocialAPI | null = getSocialClient(store.getState())
   if (!client) return
 
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -461,7 +461,7 @@ export function getFriendsWithDirectMessages(request: GetFriendsWithDirectMessag
     request.userNameOrId
   )
 
-  const friendsConversations: Array<{ userId: string; conversation: Conversation }> = []
+  const friendsConversations: Array<{ userId: string; conversation: Conversation; avatar: Avatar }> = []
 
   for (const friend of filteredFriends) {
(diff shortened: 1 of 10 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -435,6 +427,70 @@ export function getFriendRequests(request: GetFriendRequestsPayload) {
...
+  const client: SocialAPI | null = getSocialClient(store.getState())
+  if (!client) return
+
+  const conversations = client.getAllCurrentConversations()
+
+  const conversationsWithMessages = conversations.filter((conv) => conv.conversation.hasMessages)
+  const unseenPrivateMessages = {}
+
+  for (const conversation of conversationsWithMessages) {
+    unseenPrivateMessages[conversation.conversation.userIds![1]] = conversation.conversation.unreadMessages?.length || 0
```
moliva, line 440: Is the index `1` always used for the other person or could it happen that it isn't?

packages/shared/friends/sagas.ts
```diff
@@ -435,6 +427,70 @@ export function getFriendRequests(request: GetFriendRequestsPayload) {
...
+  getUnityInstance().UpdateTotalUnseenMessagesByUser(totalUnseenMessages)
+}
+
+export function getFriendsWithDirectMessages(request: GetFriendsWithDirectMessagesPayload) {
+  const client: SocialAPI | null = getSocialClient(store.getState())
+  if (!client) return
+
+  const conversations = client.getAllCurrentConversations()
+
+  const conversationsWithMessages = conversations.filter((conv) => conv.conversation.hasMessages)
```
moliva, line 455: How about extracting this into a selector function as it seems it's a common pattern (also used in `getUnseenMessagesByUser`)?

packages/shared/friends/sagas.ts
jmoguilevsky: It's always the other person https://github.com/decentraland/matrix-client/blob/0d3614621c8020201da2e75b46b2daf9aff18ebe/src/MessagingClient.ts#L65

moliva (approved):

PR #388, opened 2022-07-28
feat: use new RPC URLs

nicosantangelo:
# What? 

It updates the infura URLs to use our new Cloudflare URLs that load balance nodes via rpc.decentraland.org
...

# Why? 
...

We're trying to minimize the amount of infura keys we have and to centralize in a single URL requests to re-destribute them later

### DIFF
--- a/static/preview.html
+++ b/static/preview.html
@@ -741,7 +741,7 @@
             kernel.authenticate(globalThis.ethereum, false)
           }
         } else {
-          kernel.authenticate('wss://mainnet.infura.io/ws/v3/f54f2e10b59647778de06d884121f8fa', true)
+          kernel.authenticate('wss://rpc.decentraland.org/mainnet', true)
         }
 
         if (isElectron()) {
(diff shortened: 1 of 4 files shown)


static/preview.html
```diff
@@ -741,7 +741,7 @@
             kernel.authenticate(globalThis.ethereum, false)
           }
         } else {
-          kernel.authenticate('wss://mainnet.infura.io/ws/v3/f54f2e10b59647778de06d884121f8fa', true)
+          kernel.authenticate('wss://rpc.decentraland.org/mainnet', true)
```
menduz, line 744: should we add the query param to identify the source app? that way we could identify the difference between "Preview app" and "localhost:5000"

nicosantangelo: Oh! Thought about that while changing the code. The URL "supports" the `project` query param to identify the source. We're using it for things like `wallet-link`, which is not the same as identifying the project itself (i.e.: `marketplace`)

Maybe we can use another querystring for this?

cc: @2fd

nicosantangelo: Don't forget about your friend nico @menduz @2fd

sorry for re-pinging but I don't want to make a change without being sure 😅

menduz: @nachomazzara will know better. I've been told to use a specific project for backends and catalysts, I think the same logic applies here

menduz: merging to unblock

PR #393, opened 2022-08-02
feat: Add support for fetching emotes

juanmahidalgo:
# What? 
We need to support fetching emotes when the `Renderer` asks for the Profile. This PRs depends heavily on the new release of the catalyst client (PR: https://github.com/decentraland/catalyst-client/pull/253). It won't build until it's released and the version is bumped here.

# Why? 
Emotes will be fetched using a different catalyst client method and will also contain a different structure so we need to process it differently. Also, the wearables that are unpublished (coming from the builder), are not retrieving the emote data correctly, so a fix is included


packages/shared/catalogs/sagas.ts
```diff
@@ -54,9 +63,39 @@ const BASE_BUILDER_DOWNLOAD_URL = `${BUILDER_SERVER_URL}/storage/contents`
  *
  */
 export function* catalogsSaga(): any {
-  yield takeEvery(WEARABLES_REQUEST, handleWearablesRequest)
+  yield takeEvery(WEARABLES_REQUEST, handleItemRequest)
   yield takeEvery(WEARABLES_SUCCESS, handleWearablesSuccess)
   yield takeEvery(WEARABLES_FAILURE, handleWearablesFailure)
+  yield takeEvery(EMOTES_REQUEST, handleItemRequest)
+  yield takeEvery(EMOTES_SUCCESS, handleWearablesSuccess)
+  yield takeEvery(EMOTES_FAILURE, handleWearablesFailure)
```
cazala, line 71: if the request handler was renamed from `wearable` to `item` should the success and failure handlers be renamed as well?

packages/shared/catalogs/sagas.ts
```diff
@@ -54,9 +63,39 @@ const BASE_BUILDER_DOWNLOAD_URL = `${BUILDER_SERVER_URL}/storage/contents`
...
+      const net: ETHEREUM_NETWORK = yield select(getSelectedNetwork)
+      const assetBundlesBaseUrl: string = getAssetBundlesBaseUrl(net) + '/'
+
+      const v2Wearables: WearableV2[] = response.map((wearable) => ({
+        ...wearable,
+        baseUrl: wearable.baseUrl ?? fetchContentServer + '/contents/',
+        baseUrlBundles: assetBundlesBaseUrl
+      }))
+
+      yield put(wearablesSuccess(v2Wearables, context))
```
cazala, line 92: this is always dispatching a `wearableSuccess` even for an `emoteRequest` (instead of an `emoteSuccess`), is this intended?

packages/shared/catalogs/sagas.ts
juanmahidalgo: absolutely, updated! :)

packages/shared/catalogs/sagas.ts
juanmahidalgo: No! It should dispatch one of the new ones, updated! thanks for the catch :)

menduz (approved):
Looks good, if tested and working: merge when you are ready.

packages/shared/catalogs/types.ts
```diff
@@ -94,6 +96,9 @@ export type PartialEmote = PartialBy<Omit<Emote, 'baseUrlBundles'>, 'baseUrl'>
 export type PartialItem = PartialWearableV2 | PartialEmote
 type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
 
+export const isPartialWearable = (partialItem: PartialItem): partialItem is PartialWearableV2 =>
+  !!(partialItem as PartialWearableV2).data
```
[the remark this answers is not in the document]
juanmahidalgo: Good question! I _believe_ it's enough since `isPartialWearable` is used to access the `data` object and the `PartialEmote` won't ever have it so adding the check sounds redundant.

cazala (approved):

cazala (approved):

gonpombo8 (approved):

PR #394, opened 2022-08-02
fix: land names and comply with DAO poll

menduz:
Removes the marketplace dependency from explorer

Complies with DAO poll https://governance.decentraland.org/proposal/?id=d43ed210-fd5f-11ec-8fbf-839e0f96469a to remove hardcoded POI list


lorux0 (dismissed):
works for me!

packages/shared/apis/host/UserActionModule.ts
```diff
@@ -28,36 +27,39 @@ export function registerUserActionModuleServiceServerImplementation(port: RpcSer
...
-          mapSceneData.entity.metadata,
-          mapSceneData.entity.content,
-          getFetchContentServer(store.getState())
-        )
+        if (!sceneData.previewImageUrl) {
+          let sceneParcels = [destination]
+          if (metadata && metadata.scene.parcels) {
+            sceneParcels = metadata.scene.parcels
+          }
+          sceneData.previewImageUrl = `https://api.decentraland.org/v1/map.png?width=480&height=237&size=10&center=${destination}&selected=${sceneParcels.join(
```
nachomazzara, line 56: Are we using this also for the dev environment? If yes, can we move this to an env var per environment?

menduz: solved

lorux0 (approved):

PR #396, opened 2022-08-03
Add support for feature flag to disable some catalysts

agusaldasoro:
Added new feature flag to list catalyst to be removed from the options in pick realm algorithm.


kuruk-mm (approved):
Great work! I tested with loki and everything is ok!

PR #401, opened 2022-08-05
feat: prefer realm selection algorithm to get people together

menduz:
If `localStorage` saved ream is chosen instead, there is no way to optimize realm selection to gather people together, instead a normal distribution across realms was achieved in the long run.

### DIFF
--- a/packages/shared/dao/sagas.ts
+++ b/packages/shared/dao/sagas.ts
@@ -70,6 +70,8 @@ export function* daoSaga(): any {
 
 function* pickCatalystRealm() {
   const candidates: Candidate[] = yield select(getAllCatalystCandidates)
+  if (candidates.length === 0) return undefined
+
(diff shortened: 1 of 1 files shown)


packages/shared/dao/sagas.ts
```diff
@@ -165,10 +167,10 @@ function* selectRealm() {
     (PREVIEW ? PREVIEW_REALM : null) ||
     // CATALYST from url parameter
     (PIN_CATALYST ? realmFromPinnedCatalyst() : null) ||
-    // cached in local storage
-    (yield call(getRealmFromLocalStorage, network)) ||
     // fetch catalysts and select one using the load balancing
-    (yield call(pickCatalystRealm))
+    (yield call(pickCatalystRealm)) ||
+    // cached in local storage
+    (yield call(getRealmFromLocalStorage, network))
```
hugoArregui, line 173: does it makes sense to have this at all? if pickCatalystRealm was unable to pick one won't that mean that the servers are down or something?

PR #402, opened 2022-08-08
fix: lazy loading chat

Julieta11:
# What? 
+ Fix typo in chat initialization. We were specifying the wrong method when sending the message to unity.
+ Update dcl-social-client dependency version in order to get the attribute `unreadMessages` when fetching all current conversations.
+ Add payload types to be able to mark as seen and get private msgs


packages/unity-interface/UnityInterface.ts
```diff
@@ -385,6 +385,7 @@ export class UnityInterface implements IUnityInterface {
   }
 
   public AddFriendsWithDirectMessages(addFriendsWithDirectMessagesPayload: AddFriendsWithDirectMessagesPayload): void {
+    console.log('AddFriendsWithDirectMessages', addFriendsWithDirectMessagesPayload)
```
jmoguilevsky, line 388:
```suggestion
```

packages/shared/friends/utils.ts
```diff
@@ -22,7 +52,27 @@ export function getUserIdFromMatrix(userId: string) {
...
-export function getMatrixIdFromUser(userId: string) {
+function getMatrixIdFromUser(userId: string) {
   // TODO check how'd be the correct way to do this
   return '@' + userId.toString() + ':decentraland.org'
 }
+
+/**
+ *
+ */
+function updateSocialInfo(socialData: SocialData) {
```
jmoguilevsky, line 63: this belongs to the sagas file, utils are just for functions that don't have any side effects

packages/shared/friends/utils.ts
```diff
@@ -14,6 +21,29 @@ export function getUserIdFromMatrix(userId: string) {
...
+    const socialId = getMatrixIdFromUser(userId)
+    const conversation: Conversation = await client.createDirectConversation(socialId)
+
+    const socialData: SocialData = {
+      userId: userId,
+      socialId: socialId,
+      conversationId: conversation.id
+    }
+
+    updateSocialInfo(socialData)
```
jmoguilevsky, line 40: same as the other function, this should be part of the sagas file

jmoguilevsky (approved):

PR #409, opened 2022-08-10
feat: add support for emotes in the Profile

juanmahidalgo:
# What? 
The profile will send `emotes` in a different array soon, we need to prepare the kernel to send to the rendered the emotes form the lambda and also send them back to the lambda when saving it.

# Why? 
This changes will be introduced with the release of Emotes, so we need to adapt the kernel so it sends them to the Renderer.


menduz (dismissed):

menduz:
merge when you ready

PR #410, opened 2022-08-10
fix: report form

nachomazzara:
Fix the deprecated forms.


menduz:
Two changes:
- Can we use dcl.gg instead?
- Linter

nachomazzara:
> Two changes:
> 
> * Can we use dcl.gg instead?
> * Linter

How can I use dcl.gg? can you point me to the doc?

menduz (approved):

PR #422, opened 2022-08-10
chore: update fetchCatalystStatuses test

hugoArregui:


test/dao/fetchCatalystStatuses.test.ts
```diff
@@ -2,37 +2,21 @@ import sinon from 'sinon'
...
   domain: 'peer.decentraland.org',
   elapsed: 309,
-  maxUsers: 1000,
+  maxUsers: -1,
   status: 0,
   usersCount: 59,
-  usersParcels: undefined
+  usersParcels: [
+    [1, 1],
+    [1, 1],
```
agusaldasoro, line 17: shouldn't the parcels be different from each other?

hugoArregui: not here, in the about response yes, but here it expects a parcel per user

agusaldasoro (approved):

menduz (approved):

PR #435, opened 2022-08-16
fix: add usersCount for loadBalancing

menduz:
# What? 
...

# Why? 
...

### DIFF
--- a/packages/shared/dao/index.ts
+++ b/packages/shared/dao/index.ts
@@ -83,7 +83,7 @@ export async function fetchCatalystStatus(
         domain: domain,
         status: aboutResponse.status,
         elapsed: aboutResponse.elapsed!,
-        usersCount: bff.userCount ?? 0,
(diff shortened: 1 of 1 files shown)


packages/shared/dao/index.ts
```diff
@@ -83,7 +83,7 @@ export async function fetchCatalystStatus(
         domain: domain,
         status: aboutResponse.status,
         elapsed: aboutResponse.elapsed!,
-        usersCount: bff.userCount ?? 0,
+        usersCount: bff.userCount ?? comms.usersCount ?? 0,
```
agusaldasoro, line 86: what happens if `bff.userCount` is `0` but `comms.usersCount` has a value, would it work?

menduz: `<lhs>??<rhs>` picks the RHS if LHS is null. it doesn't fall into RHS if falsy.


Only will pick `comms.usersCount` if `bff.userCount` is null or undefined

PR #442, opened 2022-08-17
feat: add channels

Julieta11:
# What? 
Base branch for channels feature. Find more info [here](https://www.notion.so/decentraland/RFC-Technical-proposal-for-Channels-8fdbd914091f4d2194424caf9e3e096b).

So far, the PR addresses: 
- join or create channel via command,
- create channel,
- ability to get all channels where the user is joined,
- and show channel messages window

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -26,11 +26,15 @@ import {
   FriendshipAction,
   PresenceStatus,
   CreateChannelPayload,
-  JoinOrCreateChannelErrorPayload,
+  ChannelErrorPayload,
   UpdateTotalUnseenMessagesByChannelPayload,
   GetJoinedChannelsPayload,
   ChannelsInfoPayload,
-  ChannelInfoPayload
+  ChannelInfoPayload,
+  MarkChannelMessagesAsSeenPayload,
+  UpdateTotalUnseenMessagesPayload,
+  GetChannelMessagesPayload,
+  AddChatMessagesPayload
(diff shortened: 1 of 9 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -799,3 +822,246 @@ function logAndTrackError(message: string, e: any) {
...
+
+  if (CHANNEL_RESERVED_IDS.includes(channelId)) {
+    joinChannelError(channelId, 'Reserved name')
+    return
+  }
+
+  if (!validateRegexChannelId(channelId)) {
+    joinChannelError(channelId, 'Bad regex')
+    return
+  }
```
moliva, line 883: Can we extract these validations and use from the previous join or create command as well? Also to unify error messages.

packages/shared/friends/sagas.ts
```diff
@@ -799,3 +822,246 @@ function logAndTrackError(message: string, e: any) {
...
+
+  // create channel
+  const { created, conversation } = await client.getOrCreateChannel(channelId, [ownId])
+
+  if (!created) {
+    joinChannelError(channelId, 'Already exists')
```
moliva, line 889:
```suggestion
    joinChannelError(channelId, `Channel "${channelId}" already exists`)
```

packages/shared/friends/sagas.ts
jmoguilevsky: Yes, this should be done from the client side

PR #450, opened 2022-08-24
Feat: adds support to save and retrieve home point

### LINKED ISSUE #444: Home point functionality
A new functionality will allow players to set an home point to let them spawn there.

The changes that are currently needed in kernel are:

- [ ]  A function that the unity renderer can call to which pass the home point coordinates (it was suggested to save this coords in the local storage and avoid saving at profile level)
- [ ] A function that the unity renderer can call to retrive the current home point (returning 0,0 if not set)
- [ ] The behavior of this should be that if in the URL there are no specific coordinates the player should spawn at the home point, otherwise the url coordinates will override this

For additional info this is the unity-renderer issue: https://github.com/decentraland/unity-renderer/issues/2021

davidejensen:
Closes #444 

Adds the required function to save and retrieve the home point coordinates,  also spawns the player in the home point position unless overwritten in the url with the "position" query parameter.

To be tested with the following unity-renderer branch:

### DIFF
--- a/packages/entryPoints/index.ts
+++ b/packages/entryPoints/index.ts
@@ -48,6 +48,7 @@ import { signalEngineReady } from 'shared/renderer/actions'
 import { IUnityInterface } from 'unity-interface/IUnityInterface'
 import { getCurrentUserProfile } from 'shared/profiles/selectors'
 import { sendHomeScene } from '../shared/atlas/actions'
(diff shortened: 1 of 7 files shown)


menduz (approved):
awesome! 🚀 
feel free to merge it if it works as expected

PR #457, opened 2022-08-26
fix: unread message count

jmoguilevsky:
# What? 
...

# Why? 
...

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -317,8 +317,7 @@ function* refreshFriends() {
 
     // init friends
     // const friends: string[] = []
-    const friends: string[] = yield client.getAllFriends()
-
+    const friendIds: string[] = yield getFriendIds(client)
     const friendsSocial: SocialData[] = []
 
     // init friend requests
@@ -346,7 +345,6 @@ function* refreshFriends() {
       {}
     )
 
-    const friendIds = friends.map(($) => parseUserId($)).filter(Boolean) as string[]
     const requestedFromIds = fromFriendRequests.map(
       (request): FriendRequest => ({
         createdAt: request.createdAt,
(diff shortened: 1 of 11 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -236,11 +248,16 @@ function* configureMatrixClient(action: SetMatrixClient) {
       recipient: message.sender === ownId ? senderUserId : identity.address
     }
 
+    let userProfile = getProfile(store.getState(), senderUserId)
+    if (!userProfile || !isAddedToCatalog(store.getState(), senderUserId)) {
+      await ensureFriendProfile(senderUserId)
```
moliva, line 253: 1) Shouldn't we send the profile to the renderer after this (as done in the previous case)?
2) If so, is there a way we can extract these two use cases of checking profile, catalog and send to renderer?

packages/shared/friends/sagas.ts
```diff
@@ -316,8 +333,7 @@ function* refreshFriends() {
 
     // init friends
     // const friends: string[] = []
```
moliva, line 335: Remove commented code.

packages/shared/friends/sagas.ts
jmoguilevsky: The function `ensureFriendProfile` sends the profile to renderer (part of the sagas dependency tree)

packages/shared/friends/sagas.ts
```diff
@@ -198,14 +200,24 @@ function* configureMatrixClient(action: SetMatrixClient) {
   }
 
   // initialize conversations
-  client.onStatusChange((socialId, status) => {
+  client.onStatusChange(async (socialId, status) => {
```
menduz, line 203: who catches errors in this function?

packages/shared/friends/sagas.ts
```diff
@@ -198,14 +200,24 @@ function* configureMatrixClient(action: SetMatrixClient) {
...
+          totalFriends: getTotalFriends(store.getState())
+        })
+      }
+
       sendUpdateUserStatus(userId, status)
     }
   })
 
-  client.onMessage((conversation, message) => {
+  client.onMessage(async (conversation, message) => {
```
menduz, line 220: who catches the errors of this async function?

packages/shared/friends/sagas.ts
jmoguilevsky: Refactored the function so that it will never throw an error

packages/shared/friends/sagas.ts
jmoguilevsky: Refactored the function so that it will never throw an error

packages/shared/friends/sagas.ts
moliva: Got it! 👍🏻

moliva (approved):
LGTM.

PR #459, opened 2022-08-26
fix: catch the evaluation error and fails gracefully

leanmendoza:
# What? 
...

# Why? 
...

### DIFF
--- a/packages/scene-system/sdk/SceneRuntime.ts
+++ b/packages/scene-system/sdk/SceneRuntime.ts
@@ -65,26 +65,10 @@ export async function startSceneRuntime(client: RpcClient) {
       `SDK: Error while loading ${url} (${mappingName} -> ${mapping?.file}:${mapping?.hash}) the mapping was not found`
     )
   }
-
   componentSerializeOpt.useBinaryTransform = explorerConfiguration.configurations['enableBinaryTransform'] === 'true'
-
-  const sourceCode = await codeRequest.text()
 
   let didStart = false
   let updateIntervalMs: number = 1000 / 30
-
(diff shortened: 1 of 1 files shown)


pravusjif (dismissed):

packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -144,36 +141,48 @@ export async function startSceneRuntime(client: RpcClient) {
...
+      })
+    }
+
+    // run the code of the scene
+    await customEval(sourceCode, runtimeExecutionContext)
+  } catch (err) {
+    await EngineAPI.sendBatch({ actions: [initMessagesFinished()] })
+
+    devToolsAdapter.error(new Error(`SceneRuntime: Error while evaluating the scene ${workerName}`))
+    await sleep(100)
```
gonpombo8, line 182: why we sleep? can we add a comment ?

packages/scene-system/sdk/SceneRuntime.ts
```diff
@@ -144,36 +142,47 @@ export async function startSceneRuntime(client: RpcClient) {
...
+
+    throw err
```
gonpombo8, line 184: can we eliminate this error since we are already catching it ?

leanmendoza: If we remove the throw, it fails silently in prod. The sleep is because the devToolsAdapter isn't an async function, if we close the worker the data can be not sent.

leanmendoza: `sleep` comment added

gonpombo8 (approved):

PR #461, opened 2022-08-31
fix: removed updating feature flag variant name

aixaCode:
# What? 
Removed updating flag variant `name` property with the same name as the feature flag.

# Why? 
To do the A/B test on the client, we need information on which variant is enabled, and the original name property was removed during this step.

Old json:
```
"test_variant": {
      "name": "test_variant",
      "payload": null,
      "enabled": true
    }
```

New json:
```
"test_variant": {
      "name": "test_variant_name_descripton", // - value passed from the feature flag config
      "payload": null,
      "enabled": true
    }
```


menduz:
Variants should have their own distinct name defined by the feature flag server, why is this necessary? Isn't having `enabled:bool` redundant?

aixaCode:
That is a good question. From my understanding enabled:bool tells if the variant is active. But agree, it's quite redundant.

menduz:
Oh I see. The clients always receive only one variant, unless the 'X-Debug: true' header is passed. I think having variants named `enabled` and `disabled` is purely an oddity in the documentation, there is also another example called `youtube` in the docs you shared. 

```bash
curl -H "x-debug: true" -H "Referer: https://play.decentraland.org" https://feature-flags.decentraland.org/explorer.json | jq
```

The `enabled:bool` field exists for the purpose of knowing whether the variant is enabled or not, but it is only there to make a compliant data type, in practice (unless using the x-debug) every received variant and FF is active.

The name of the variant was meant to identify the cohort of the experiment, like the group A or B of the A/B testing. The key of the object containing the variant contains the feature flag name. The datatype [are defined here ](https://github.com/decentraland/feature-flags/blob/main/src/index.ts#L16-L39)

aixaCode:
You are right, I will make the changes and update the PR. Thanks for feedback

aixaCode:
@menduz Made the changes, for consistency, in the debugging options, if a variant exists, the feature flag will be enabled.

menduz (approved):

menduz:
Thanks for the fixes! Feel free to merge when needed

PR #469, opened 2022-09-01
Fix:  generate default avatars for invalid/inexistent users

jmoguilevsky:
# What? 
...

# Why? 
...


packages/shared/profiles/sagas.ts
```diff
@@ -268,7 +273,15 @@ function processRemoteProfiles(profiles: RemoteProfile[], _userIds: string[]): A
     })
     .filter((avatar: Avatar | null): boolean => avatar !== null) as Array<Avatar>
 
-  return avatars
+  // come up with a random profile for users that don't have one
```
jmoguilevsky, line 276:
```suggestion
  // generate a random profile for users that don't have one
```

moliva (approved):
LGTM!

PR #474, opened 2022-09-05
send home point always to renderer

gonpombo8:
# What? 
...

# Why? 
...


davidejensen:
Works as expected, thanks!

davidejensen (approved):

PR #478, opened 2022-09-05
feat: extracting profiles as promise feat to a separate branch

moliva:
Part of #363.

# What? 
Introduces new profiles batch endpoint.

# Why? 
...


packages/shared/profiles/sagas.ts
```diff
@@ -150,19 +155,16 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
 
     const profile: Avatar =
       // first fetch avatar through comms
       (shouldFetchViaComms && (yield call(requestProfileToPeers, commsContext, userId, version))) ||
       // and then via catalyst
       (shouldLoadFromCatalyst && (yield call(getRemoteProfile, userId, version))) ||
       // then for my profile, try localStorage
-      (shouldReadProfileFromLocalStorage && (yield call(readProfileFromLocalStorage))) ||
-      // lastly, come up with a random profile
-      (shouldFallbackToRandomProfile && (yield call(generateRandomUserProfile, userId)))
```
menduz, line 163: Where does this fallback take place?

packages/shared/profiles/sagas.ts
```diff
@@ -183,33 +185,112 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
   }
 }
 
-function* getRemoteProfile(userId: string, version?: number) {
-  try {
-    const profiles: { avatars: Avatar[] } = yield call(profileServerRequest, userId, version)
+export function* handleFetchProfiles(action: ProfilesRequestAction): any {
```
menduz, line 188: Why does this function called fetchProfiles tries to load from localStorage?

packages/shared/profiles/sagas.ts
```diff
@@ -183,33 +185,112 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
+  if (!identity) throw new Error("Can't fetch profile if there is no ExplorerIdentity")
 
-      avatar.hasClaimedName = !!avatar.name && avatar.hasClaimedName // old lambdas profiles don't have claimed names if they don't have the "name" property
-      avatar.hasConnectedWeb3 = true
+  try {
+    const userIdsToFetch = userIds.filter((userId) => userId.toLowerCase() !== identity.address.toLowerCase())
+    const isFetchingOwnUser = userIds.some((userId) => userId.toLowerCase() === identity.address.toLowerCase())
 
-      return avatar
+    const avatars: Avatar[] = yield (!isFetchingOwnUser && call(getRemoteProfiles, userIdsToFetch)) || []
```
menduz, line 199:
```suggestion
    const avatars: Avatar[] = isFetchingOwnUser ? (yield call(getRemoteProfiles, userIdsToFetch) || []) : []
```

packages/shared/profiles/sagas.ts
```diff
@@ -183,33 +185,112 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
+}
+
+function* getRemoteProfile(
+  userId: string,
+  version?: number
+): Generator<CallEffect<RemoteProfile> | CallEffect<Array<Avatar>> | RemoteProfile> {
+  try {
+    const remoteProfile: RemoteProfile = (yield call(profileServerRequest, userId, version)) as RemoteProfile
+
+    const profiles: Avatar[] = (yield call(processRemoteProfiles, [remoteProfile], [userId])) as Array<Avatar>
```
menduz, line 224:
```suggestion
    const profiles: Avatar[] = (yield call(processRemoteProfiles, [remoteProfile], [userId]))
```

packages/shared/profiles/sagas.ts
```diff
@@ -183,33 +185,112 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
+    return profiles[0]
+  } catch (error: any) {
+    if (error.message !== 'Profiles not found') {
       defaultLogger.log(`Error requesting profile for auth check ${userId}, `, error)
     }
   }
   return null
```
menduz, line 231: The code is not correct but something like this should be applied
```suggestion
  const results: Array<Avatar | undeined> = yield call(getRemoteProfiles, [userId])
  return results[0]
```

packages/shared/profiles/sagas.ts
```diff
@@ -440,7 +546,7 @@ async function generateRandomUserProfile(userId: string): Promise<Avatar> {
 
   let profile: Avatar | undefined = undefined
   try {
-    const profiles: { avatars: Avatar[] } = await profileServerRequest(`default${_number}`)
+    const profiles: RemoteProfile = await profilesServerRequest([`default${_number}`])[0]
```
menduz, line 549:
```suggestion
    const profiles: RemoteProfile = await profileServerRequest(`default${_number}`)
```

packages/unity-interface/BrowserInterface.ts
```diff
@@ -600,7 +600,7 @@ export class BrowserInterface {
     if (message.action === FriendshipAction.REQUESTED_TO) {
       const avatar = await ensureFriendProfile(userId)
 
-      if (isAddress(userId)) {
+      if (isAddress(userId) && avatar) {
```
menduz, line 603: This is contractually wrong, the `ensure` function should always return a valid avatar

menduz (changes requested):

packages/shared/profiles/sagas.ts
moliva: [Here](https://github.com/decentraland/kernel/pull/478/files#diff-0119c0e28687b0c42257febe4d02be417a56f511d1582182c47f25076871bb3dR285).

packages/unity-interface/BrowserInterface.ts
moliva: It seems this function can fail and is now catched, returning a lack of value. Will remove the catch from the function and let another scope handle it.

packages/shared/profiles/sagas.ts
moliva: This does not account for the user profile `version` that the single fetch profile one does.

packages/shared/profiles/sagas.ts
menduz: can this be moved back here? otherwise the logic to resolve profiles is hard to follow deep inside the stack trace. it is not clear enough that `getRemoteProfile` generates a profile if it is not found

packages/shared/friends/ensureFriendProfile.ts
```diff
@@ -1,22 +1,9 @@
...
-    const message = 'Failed while ensuring friend profile'
-    defaultLogger.error(message, error)
-
-    trackEvent('error', {
-      context: 'kernel#saga',
-      message: message,
-      stack: '' + error
-    })
-  }
+export async function ensureFriendProfile(userId: string): Promise<Avatar> {
```
menduz, line 5: I think that ensureFriendProfile is a better name for ProfileAsPromise if you want to rename it. Otherwise this two functions should be deleted

packages/shared/profiles/ProfileAsPromise.ts
```diff
@@ -1,25 +1,36 @@
...
   const existingProfileWithCorrectVersion = existingProfile && isExpectedVersion(existingProfile)
-  if (existingProfile && existingProfileWithCorrectVersion) {
-    return Promise.resolve(existingProfile)
+  const isInCatalog = isAddedToCatalog(store.getState(), userId)
+  if (existingProfile && existingProfileWithCorrectVersion && !isCurrentUserId(store.getState(), userId)) {
+    if (!isInCatalog) {
+      await ensureUnityInterface()
+      getUnityInstance().AddUserProfileToCatalog(profileToRendererFormat(existingProfile, {}))
+      store.dispatch(addedProfileToCatalog(userId, existingProfile))
+    }
```
menduz, line 32: All of this should not happen inside this function

menduz: saga forwardProfileToRenderer

packages/shared/friends/ensureFriendProfile.ts
moliva: In the spirit of breaking a big PR into smaller pieces, how about we leave this change for a very small refactor PR with only this change in the week?

packages/shared/profiles/actions.ts
```diff
@@ -20,6 +24,9 @@ export const SEND_PROFILE_TO_RENDERER = 'SEND_PROFILE_TO_RENDERER'
 export const profileRequest = (userId: string, profileType?: ProfileType, version?: number) =>
   action(PROFILE_REQUEST, { userId, profileType, version })
```
menduz, line 25: I'd remove this one keeping only `profilesRequest([singleUserId])`

packages/shared/profiles/sagas.ts
```diff
@@ -98,7 +103,7 @@ function* initialRemoteProfileLoad() {
   let profile: Avatar
 
   try {
-    profile = yield call(ProfileAsPromise, userId, undefined, isGuest ? ProfileType.LOCAL : ProfileType.DEPLOYED)
+    profile = yield call(ensureFriendProfile, userId, isGuest ? ProfileType.LOCAL : ProfileType.DEPLOYED)
```
menduz, line 106: This one now breaks the logic now, rendering all the fallbacks below unreachable because the function generates a random profile if it can't be loaded from the server

moliva:
Closing this PR FTTB.

PR #484, opened 2022-09-07
feat: friends lazy load

moliva:
# What? 
- Implementing friends and direct messages lazy loading
- Removes logic from previous PR with profiles batch endpoint

# Why? 
Make friends and private messaging a scalable solution and improve UX.


packages/shared/profiles/ProfileAsPromise.ts
```diff
@@ -1,25 +1,28 @@
...
   function isExpectedVersion(aProfile: Avatar) {
     return !version || aProfile.version >= version
   }
 
   const [, existingProfile] = getProfileStatusAndData(store.getState(), userId)
   const existingProfileWithCorrectVersion = existingProfile && isExpectedVersion(existingProfile)
-  if (existingProfile && existingProfileWithCorrectVersion) {
-    return Promise.resolve(existingProfile)
+  if (existingProfile && existingProfileWithCorrectVersion && !isCurrentUserId(store.getState(), userId)) {
+    store.dispatch(profileSuccess(existingProfile))
```
menduz, line 24: in which case this line is necessary?

moliva: This makes sure we go through the process posterior to a profile being fetched, particularly forwarding it to the renderer.

packages/shared/profiles/sagas.ts
```diff
@@ -183,33 +183,67 @@ export function* handleFetchProfile(action: ProfileRequestAction): any {
...
+    })
+    .filter((avatar: Avatar | null): boolean => avatar !== null) as Array<Avatar>
+
+  // generate a random profile for users that don't have one
+  // this is so that renderer can show them to the user
+  // and show them in the chat so no ghost notification is left hanging
+  const usersWithoutAvatars = userIds.filter((userId) => !avatars.some((avatar) => avatar.userId === userId))
+  const defaultAvatars = await Promise.all(
+    usersWithoutAvatars.map(async (userId): Promise<Avatar> => generateRandomUserProfile(userId))
+  )
```
menduz, line 241: I think this should not be part of the logic of `processRemoteProfiles` because it alters the initial loading of the "self" profile

packages/shared/profiles/ProfileAsPromise.ts
menduz: in which cases does a profile exist locally and has not been sent to the renderer? I'm afraid that if we merge this line the next refactor removes it and breaks an use case

packages/shared/profiles/sagas.ts
moliva: Sorry, this is a leftover from a rollbacked change. Will remove it now.

packages/shared/profiles/ProfileAsPromise.ts
moliva: I think the new version is what we are actually trying to do here, not trigger a new action, but rather making sure that the profile is added to the catalog before this promise resolves.

If we do not want this behavior, please, let's sync and find an alternative solution.

menduz (dismissed):
Looking good from a code PoV! Didn't test the complete flow, merge it when you need it

menduz (approved):

PR #487, opened 2022-09-08
feat: Remove version field from the profile's avatar object

juanmahidalgo:
# What? 
Since the Renderer won't need this field anymore, this PR removes it from the `avatar` object. This is a follow-up PR to: https://github.com/decentraland/kernel/pull/486.

# Why? 
It was passed to the Renderer so it knows when to send emotes separately or not but since it will always send them as `emotes`, it's not needed.


menduz (approved):
🚢 it

PR #505, opened 2022-09-13
chore: support emotes in avatarshape

menduz:


leanmendoza (approved):

packages/ui/avatar/avatarSystem.ts
```diff
@@ -76,6 +76,7 @@ export class AvatarEntity extends Entity {
       // this should be deleted once this PR is merged and implemented in renderer
       // https://github.com/decentraland/protocol/pull/39
       if (profile.avatar.emotes) {
+        shape.emotes = profile.avatar.emotes.map(($) => $.urn)
```
cazala, line 79: is it okay that the slots are not being sent?

menduz: slots are part of the local profile, the SDK components should be as simple as possible

PR #527, opened 2022-09-19
feat: add validation on how many channels a user can join

### LINKED ISSUE #518: Add client validations for limiting channels
Add a limit to the channels a user can join in total. Make this number configurable through feature flags.

This should be checked on join and join/create.
Define error message with Explorer team.

Julieta11:
Closes #518 

# What? 
Check with a feature flag if the user has reached the maximum allowed number of channels they can join.


gonpombo8 (approved):

packages/shared/friends/sagas.ts
```diff
@@ -1197,10 +1197,17 @@ function updateSocialInfo(socialData: SocialData) {
 // Join or create channel
 async function* handleJoinOrCreateChannel(action: JoinOrCreateChannel) {
```
menduz, line 1198: We may want to rebase this for `async function*` -> `function *`

menduz (dismissed):

menduz (approved):

PR #536, opened 2022-09-23
feat: ab detection functionality

Antivortex:
# What? 

This PR adds a new chat command to send a message to Unity to detect which shapes are loaded as ABs and which as raw GLTFs. Command supports arguments to enable/disable shape coloring and to make the coloring local to current scene.

It is supposed to be tested together with Renderer PR:
https://github.com/decentraland/unity-renderer/pull/3099
Which closes this issue:
https://github.com/decentraland/unity-renderer/issues/2911

New chat commands are:

/detectabs => enable for all shapes
/detectabs off => disable for all shapes
/detectabs scene => enable for current scene shapes
/detectabs scene off => disable for current scene shapes (works only if "/decetabs scene" was used for this scene to enable) 

# Why? 

We need this debug tool in order to easily see if the shape was not converted to AB for some reason and also it will be simpler to understand how fast the conversion happens.

### DIFF
--- a/packages/shared/chat/sagas.ts
+++ b/packages/shared/chat/sagas.ts
@@ -242,10 +242,21 @@ function initChatCommands() {
     }
   })
 
-  addChatCommand('debug', 'Show debug panel', (_message) => getDebugPanelMessage())
+  addChatCommand('debug', 
+      'Show debug panel', 
+      (_message) => getDebugPanelMessage())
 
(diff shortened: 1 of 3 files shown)


menduz (approved):

menduz:
Looking good! Please run `make lint-fix` and merge when you consider opportunistic

kuruk-mm (dismissed):

menduz (approved):

PR #547, opened 2022-09-27
fix: lazy-loading fixes

### LINKED ISSUE #541: Fix lazy-loading related issues
- [X] For `totalFriendsWithDirectMessages` we're sending the same value as the requested one, instead of the actual value of the total friends with DMs.

- [x]  For `totalReceivedFriendRequests` and `totalSentFriendRequests` we're sending the same value as the requested one, instead of the actual value of total requests.

- [x] Pagination does not work in `AddFriendsWithDirectMessages`.

Julieta11:
Closes #541 

# What? 

- [x]  For `totalFriendsWithDirectMessages` we're sending the same value as the requested one, instead of the actual value of the total friends with DMs.

- [x]  For `totalReceivedFriendRequests` and `totalSentFriendRequests` we're sending the same value as the requested one, instead of the actual value of total requests.

- [x] Pagination does not work in `AddFriendsWithDirectMessages`.

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -206,8 +206,10 @@ function* configureMatrixClient(action: SetMatrixClient) {
       if (userId) {
         // When it's a friend and is not added to catalog
         // unity needs to know this information to show that the user has connected
-        if (isFriend(store.getState(), userId) && !isAddedToCatalog(store.getState(), userId)) {
+        if (!isAddedToCatalog(store.getState(), userId)) {
           await ensureFriendProfile(userId)
+        }
+        if (isFriend(store.getState(), userId)) {
           getUnityInstance().AddFriends({
             friends: [userId],
             totalFriends: getTotalFriends(store.getState())
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -748,12 +748,22 @@ function* initializeStatusUpdateInterval() {
       continue
     }
 
-    const domain = client.getDomain()
-
     const rawFriends: string[] = yield select(getPrivateMessagingFriends)
 
     const friends = rawFriends.map((x) => {
-      return `@${x}:${domain}`
+      return getMatrixIdFromUser(x)
+    })
+
+    // When it's a friend and is not added to catalog
+    // unity needs to know this information to show that the user has connected
+    friends.forEach(async (friend) => {
+      if (!isAddedToCatalog(store.getState(), friend)) {
+        await ensureFriendProfile(friend)
+      }
+      getUnityInstance().AddFriends({
+        friends: [friend],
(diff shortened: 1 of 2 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -220,8 +220,10 @@ function* configureMatrixClient(action: SetMatrixClient) {
       if (userId) {
         // When it's a friend and is not added to catalog
         // unity needs to know this information to show that the user has connected
-        if (isFriend(store.getState(), userId) && !isAddedToCatalog(store.getState(), userId)) {
+        if (!isAddedToCatalog(store.getState(), userId)) {
           await ensureFriendProfile(userId)
+        }
+        if (isFriend(store.getState(), userId)) {
```
moliva, line 226: I'm wondering about keeping this block the same as before instead. Two questions related to this:
1- What's the value of sending the profile to renderer if it's not a friend (at least for the logic here)?
2- Is it idempotent for renderer to `AddFriends` a given user id?

packages/shared/friends/sagas.ts
```diff
@@ -762,12 +764,22 @@ function* initializeStatusUpdateInterval() {
...
+    // When it's a friend and is not added to catalog
+    // unity needs to know this information to show that the user has connected
+    friends.forEach(async (friend) => {
+      if (!isAddedToCatalog(store.getState(), friend)) {
+        await ensureFriendProfile(friend)
+      }
+      getUnityInstance().AddFriends({
+        friends: [friend],
+        totalFriends: getTotalFriends(store.getState())
+      })
```
moliva, line 782: Fine with this, but extract the arrow function into a function to be used here and above (in the previous comment) for ensuring a friend profile is sent to renderer and added as friend as well.

moliva (commented):
Leaving some comments.

packages/shared/friends/sagas.ts
Julieta11: Comment discarded because I've reverted this change.

packages/shared/friends/sagas.ts
Julieta11: 1- Yes, you're right. It doesn't have any. So, instead, I changed the code to: 
```
if (isFriend(store.getState(), userId)) {
      if (!isAddedToCatalog(store.getState(), userId)) {
        await ensureFriendProfile(userId)
      }
      getUnityInstance().AddFriends({
        friends: [userId],
        totalFriends: getTotalFriends(store.getState())
      })
}
```
Because there is a scenario where we already sent the profile but not the friendship info.

2- Yes.

moliva: Great, thank you!

menduz (approved):

guidota (approved):

PR #549, opened 2022-09-28
feat: change io audio devices 

### LINKED ISSUE #544: Change input/output audio sources via unity-renderer
## Description
Unity-renderer doesn't have a built-in tool for changing audio devices. Thus, we need to do it via a web browser, for which we need to set up communication messages between the renderer and kernel. 

## Acceptence criteria
Renderer can 
* [ ] get available input/output audio devices from the kernel
* [ ] set chosen input/output audio devices to the kernel

popuz:
Closes #544 

# What?
Possibility for the user to change input audio device from the Renderer UI.

# Why?
Users can choose which audio device to use for speaking in DCL


packages/shared/voiceChat/sagas.ts
```diff
@@ -201,7 +205,7 @@ function* requestUserMedia() {
   const voiceHandler: VoiceHandler | null = yield select(getVoiceHandler)
   if (voiceHandler) {
     if (!voiceHandler.hasInput()) {
-      const media = yield call(requestMediaDevice)
+      const media = yield call(requestMediaDevice) // TODO(Mateo): use inputDevieId
```
kuruk-mm, line 208:
```suggestion
      const media = yield call(requestMediaDevice)
```

kuruk-mm (approved):
Great work!

kuruk-mm (approved):

PR #561, opened 2022-10-03
Update comms protocol to remove time as part of the transport message

hugoArregui:


packages/shared/comms/peers.ts
```diff
@@ -64,8 +64,8 @@ export function setupPeer(address: string): PeerInformation {
   if (!peerMap.has(ethereumAddress)) {
     const peer: PeerInformation = {
       ethereumAddress,
-      lastPositionUpdate: 0,
-      lastProfileUpdate: 0,
+      lastPositionIndex: 0,
+      lastProfileVersion: -1,
```
menduz, line 68: this is uint in the protocol
```suggestion
      lastProfileVersion: 0,
```

menduz (approved):

packages/shared/comms/peers.ts
hugoArregui: yes, but this way a profile message with version zero triggers an update, we are not sending this

PR #562, opened 2022-10-03
fix: add wait for loading screen on teleport

dalkia:
Fixes [#1615](https://github.com/decentraland/unity-renderer/issues/1615)

# What? 

Adds one extra call with renderer, in which it wais for the loading hud to be visible before making the call to initiate the teleport.

# Why? 

This is necessary because we are delegating the responsibility of showing the initial loading screen to the renderer. In this way, no unload starts until the loading hud is visible

# IMPORTANT NOTE

Don't merge until the renderer [PR](https://github.com/decentraland/unity-renderer/pull/3164) has been released


menduz (approved):
Merge after renderer release

PR #567, opened 2022-10-04
fix: dedup friend status updates

### LINKED ISSUE #504: Dedup status updates per user
[Kernel is sending a big number of user presence updates](https://grafana.decentraland.systems/d/K-UJF8z4z/miscellaneous-kernel-renderer-calls-metrics?orgId=1&var-counterName=All&from=now-7d&to=now) to renderer with duplicated state.

We need to understand if Matrix is duplicating the presence updates and if so understand how to simplify this from the producer side or possibly find a way to avoid sending duplicated data to the renderer.

lauti7:
Closes #504 

# What? 
This PR adds a Map to `Friends State` in order to cache the last status of the user's friends sent to the unity renderer

# Why? 
Avoid sending duplicate status to the renderer and decrease the number of messages sent.


packages/shared/friends/sagas.ts
```diff
@@ -480,7 +481,9 @@ function* refreshFriends() {
         socialInfo,
         friends: friendIds,
         fromFriendRequests: requestedFromIds,
-        toFriendRequests: requestedToIds
+        toFriendRequests: requestedToIds,
+        // initialize an empty map becuase there is no way to get the current statuses, the matrix client store is empty at this point
```
moliva, line 485: Typo:
```suggestion
        // initialize an empty map because there is no way to get the current statuses, the matrix client store is empty at this point
```

packages/shared/friends/sagas.ts
```diff
@@ -796,9 +799,17 @@ function sendUpdateUserStatus(id: string, status: CurrentUserStatus) {
...
 
   statuses.forEach((value: CurrentUserStatus, key: string) => {
-    sendUpdateUserStatus(key, value)
+    const lastStatus = lastStatuses.get(key)
+    if (!lastStatus || !deepEqual(lastStatus, value)) {
+      sendUpdateUserStatus(key, value)
+      lastStatuses.set(key, value)
+    } else {
+      // avoid sending already sent states
+      logger.log(`friend ${key} status was ignored - duplicated state`)
```
moliva, line 811: Let's set this to `trace` instead:
```suggestion
      logger.trace(`friend ${key} status was ignored - duplicated state`)
```

moliva (commented):
Leaving a couple of comments. Plus, let's add some test cases for this new feat.

packages/shared/friends/sagas.ts
menduz: I'd remove the trace/log from the PR before merging, to reduce the noise on the console and to prevent memory errors of these logs which in normal sessions are never garbage collected

lauti7:
> Leaving a couple of comments. Plus, let's add some test cases for this new feat.

Tests done!

packages/shared/friends/sagas.ts
lauti7: Didn't see this comment. Okay, understood! Just to know: In which sessions are the logs garbage collected?

menduz: That only happens if you clear the logs. Or depends entirely on the browser if the console is disabled.

For that reason we add DEBUG_COMMS to enable the comms logger as an example

moliva (commented):
Need to drop the `only` in the tests for all the suite to run, otherwise all good on my side.

moliva (approved):
LGTM.

menduz (approved):

PR #568, opened 2022-10-04
test: add scene number - pravus patch

pravusjif:
hacked to fix sending scene number in WSS load and unload scene messages


packages/shared/world/SceneWorker.ts
```diff
@@ -182,7 +184,7 @@ export class SceneWorker {
       this.ready |= SceneWorkerReadyState.DISPOSED
     }
 
-    getUnityInstance().UnloadScene(this.loadableScene.id)
+    getUnityInstance().UnloadScene(this.loadableScene.sceneNumber)
```
pravusjif, line 187: How should we keep this kernel update retro-compatible with the `dev` branch renderer using scene id? ...

PR #574, opened 2022-10-06
feat: implement ADR-81 comms semantics 

menduz:
Implements ADR-81 and RFC-4 & RFC-5 with comms v3

# Pending tasks
- [ ] Write tests for `IBff` sagas
- [ ] Write tests for Comms Adapter sagas
- Check https://github.com/decentraland/comms-v3/issues/45
- [ ] Ensure that loading screen semantics are on check and tested
- [x] `/changerealm menduz.dcl.eth` should unsettle the position, load the scene of menduz.dcl.eth and upon SceneReady message, settle the position and teleport myself to spawn point

# QA & Exit criteria

\* must work: means that all comms functions work: view peers, chat, voice chat, seeing people already connected and vice versa

### Loading
- [ ] Loading kernel with `realm=alsjdfhoaisudyf0a78sdyf` should fallback to a DAO realm
- [ ] Loading kernel with `realm=menduz.dcl.eth` must work and teleport the user to the scene position
- [ ] Loading kernel with `realm=worlds-content-server.decentraland.zone/world/Mariano.dcl.eth` must work and teleport to the scene position
- [ ] Loading in DAO realm and executing `/changerealm offline?cityLoaderContentServer=` should unload the world
- [ ] Loading `realm=nonexistentent.dcl.eth` should fallback to a DAO realm
- [ ] `/changerealm offline` should disconnect comms and keep the genesis city of mainnet

### Commands

- [ ] Inside `menduz.dcl.eth` and executing `/changerealm baldr` must connect to the DAO catalyst
- [ ] Inside `baldr` and executing `/changerealm menduz.dcl.eth` must connect to the preview for that ENS and teleport the user to that position
- [ ] `/changerealm <dao name>` must change the realm seamlessly
- [ ] `/changerealm INVALID_NAME` must fail, keeping the current connection open
- [ ] `/changerealm https://peer.decentraland.org` must work (load balanced)
- [ ] `/changerealm https://decentraland.org` fails
- [ ] Denylisted realms should never be selected unless `&CATALYST=` is specified
- [ ] Comms with the new CLI must work, a welcome message must be shown
- [ ] Changing from `ws-room` to `lighthouse` based peers must work
- [ ] Refreshing in without query parameters should connect to a DAO catalyst
- [ ] Specifying `&CATALYST=` should force that server bundle
- [ ] The whole comms connection should happen withing the first second of seeing the loading screen
- [ ] (v2) Joining in a crowded space should print an island, walking away should migrate us to a different island. Comming back should reconnect our explorer to the island
- [x] Changing realms create many audio worklets

# QA for v3 (not required for this PR)
- [ ] `/changerealm https://peer.decentraland.zone` must work (v3 p2p)
- [ ] Comms v3 P2P should change the `island=` query parameter in the URL when changing islands
- [ ] `/changerealm https://peer-ec2.decentraland.org` must work (room service)

# Issues blocking comms V3 but not this merge
- [ ] `Mesh` should be an instance per BFF and not per Island to preserve connections (review)


pravusjif:
There are a couple of things still not working correctly in this PR:
* the `position` query param doesn't work, it always gets overriden to spawn at 0,0, this can be tested entering https://play.decentraland.zone/?kernel-branch=adr-81-wip&position=26,-118
* the `LOS` query param also doesn't work

pravusjif:
* https://play.decentraland.org/?kernel-branch=adr-81-wip&position=100,100&LOS=1 still overrides the position and takes me to 0,0
* https://play.decentraland.org/?kernel-branch=adr-81-wip&position=100,100&LOS=0 never activates the rendering (notice I changed `LOS=1` to `LOS=0`)

pravusjif:
* `position` query param is fixed!
* `LOS` query param still has issues: right now it's working as "LOS-1" because if we use `LOS=1` the real scene load radius being mistakenly applied is 0, if we use `LOS=2` the real one applied is 1. Maybe that's why using `LOS=0` the world never starts.
* Apart from that, if we use the `LOS` query param, then the SceneLoadRadius setting  doesn't work in-world, this was particularily useful for debugging, starting the world with a specifig `LOS` and them modifying it in-world afterwards.
* This probably is not the moment, but the "LOS" from "Line Of Sight" never made much sense, eventually we might want to rename/add `SLR` from "Scene Load Radius".

PR #580, opened 2022-10-07
fix: send partial profiles to explorer to show up chat windows

guidota:
# What? 
1. Sends partial profiles to the explorer side when retrieving channel messages or members that are not in the catalog.
Partial profiles mean default avatars with a display name and image-related data to show in the chat window.

2. Prevent the creation of a channel in case a user who does not have permission (`users_allowed_to_create_channels`) to create it executes the /join command.

# Why? 
1. When opening a channel chat window or retrieving members, some of them are probably not in the catalog, and loading them there is very expensive and makes the experience to be bad.

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -103,7 +103,7 @@ import { SET_WORLD_CONTEXT } from 'shared/comms/actions'
 import { getRealm } from 'shared/comms/selectors'
 import { Avatar, EthAddress } from '@dcl/schemas'
 import { trackEvent } from '../analytics'
-import { getCurrentIdentity, getIsGuestLogin } from 'shared/session/selectors'
+import { getCurrentIdentity, getCurrentUserId, getIsGuestLogin } from 'shared/session/selectors'
 import { store } from 'shared/store/isolatedStore'
 import { getPeer } from 'shared/comms/peers'
 import { waitForMetaConfigurationInitialization } from 'shared/meta/sagas'
@@ -115,13 +115,15 @@ import {
   getMatrixIdFromUser,
   areChannelsEnabled,
   getMaxChannels,
-  getNormalizedRoomName
+  getNormalizedRoomName,
+  getUsersAllowedToCreate
 } from './utils'
 import { AuthChain } from '@dcl/kernel-interface/dist/dcl-crypto'
 import { mutePlayers, unmutePlayers } from 'shared/social/actions'
 import { getFetchContentUrlPrefix } from 'shared/dao/selectors'
 import { NewProfileForRenderer } from 'shared/profiles/transformations/types'
 import { calculateDisplayName } from 'shared/profiles/transformations/processServerProfile'
+import { uuid } from 'atomicHelpers/math'
 
 const logger = DEBUG_KERNEL_LOG ? createLogger('chat: ') : createDummyLogger()
 
@@ -726,6 +728,8 @@ export async function getPrivateMessages(getPrivateMessagesPayload: GetPrivateMe
     initialSize: limit,
     limit
   })
+
+  if (!cursorMessage) return
 
   const messages = cursorMessage.getMessages()
   if (messageId !== undefined) {
@@ -1378,27 +1382,46 @@ function* handleJoinOrCreateChannel(action: JoinOrCreateChannel) {
       return
     }
 
-    // get or create channel
-    const { created, conversation }: GetOrCreateConversationResponse = yield apply(client, client.getOrCreateChannel, [
-      channelId,
-      []
-    ])
+    // check if the user has perms to create channels.
+    const isAllowed = isAllowedToCreate()
+    if (isAllowed) {
+      const { created, conversation }: GetOrCreateConversationResponse = yield apply(
+        client,
+        client.getOrCreateChannel,
+        [channelId, []]
+      )
 
-    const channel: ChannelInfoPayload = {
-      name: conversation.name ?? action.payload.channelId,
-      channelId: conversation.id,
-      unseenMessages: 0,
-      lastMessageTimestamp: undefined,
-      memberCount: 1,
-      description: '',
-      joined: true,
-      muted: false
-    }
+      const channel: ChannelInfoPayload = {
+        name: conversation.name ?? action.payload.channelId,
+        channelId: conversation.id,
+        unseenMessages: 0,
+        lastMessageTimestamp: undefined,
+        memberCount: 1,
+        description: '',
+        joined: true,
+        muted: false
+      }
 
-    if (created) {
-      getUnityInstance().JoinChannelConfirmation({ channelInfoPayload: [channel] })
+      if (created) {
+        getUnityInstance().JoinChannelConfirmation({ channelInfoPayload: [channel] })
+      } else {
+        yield apply(client, client.joinChannel, [conversation.id])
+      }
+      // if the user does not have perms to create, we check if the channel exists and join if so.
     } else {
-      yield apply(client, client.joinChannel, [conversation.id])
+      const channelByName = yield apply(client, client.getChannelByName, [channelId])
+
+      if (channelByName) {
+        yield apply(client, client.joinChannel, [channelByName.id])
+      } else {
+        getUnityInstance().AddMessageToChatWindow({
+          messageType: ChatMessageType.SYSTEM,
+          messageId: uuid(),
+          sender: 'Decentraland',
+          body: `Ups, sorry! It seems you don't have permissions to create a channel.`,
+          timestamp: Date.now()
+        })
+      }
     }
   } catch (e) {
     if (e instanceof ChannelsError) {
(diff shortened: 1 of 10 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -1370,10 +1384,14 @@ function* handleJoinOrCreateChannel(action: JoinOrCreateChannel) {
     }
 
     // get or create channel
-    const { created, conversation } = yield apply(client, client.getOrCreateChannel, [channelId, []])
+    const { created, conversation }: { created: boolean; conversation: Conversation } = yield apply(
```
moliva, line 1387: nit: Do we have a better type name here to use?

packages/shared/friends/sagas.ts
```diff
@@ -1370,10 +1384,14 @@ function* handleJoinOrCreateChannel(action: JoinOrCreateChannel) {
...
+      client,
+      client.getOrCreateChannel,
+      [channelId, []]
+    )
 
     const channel: ChannelInfoPayload = {
-      name: action.payload.channelId,
+      name: conversation.name ?? action.payload.channelId,
```
moliva, line 1394: Should we have the `channelId` by default or remove it completely and ensure that `conversation.name` is present?

packages/shared/friends/sagas.ts
```diff
@@ -456,6 +459,17 @@ function* initializePrivateMessaging() {
     }
   ])
 
+  const profile: Avatar | null = yield select(getCurrentUserProfile)
+  if (profile) {
```
menduz, line 463: Why is this necessary? The renderer already both 1) knows our profile pic and 2) can generate it from scratch

guidota: When a user logs in, we are sending the name and image hash to Matrix, so other users that don't know us can have that information (so we don't resolve profiles as always)

menduz: can we prevent needing this?... unity has an internal texture called `AvatarTexture(address)` that can be used at any time to get the texture based on an address. using that may remove the need of passing that information at all

guidota: I believe we can, but what exactly do you want to prevent:

- hashing the content
- send this information to Matrix
- send image hash
- send default profiles from kernel to the explorer

or all of them?

sending just the address will imply some handling on the Explorer side, cc @sandrade-dcl @lorux0, are you OK with that?

sandrade-dcl: hmmm... I don't know any `AvatarTexture(address)` in the Unity side. @menduz maybe you're meaning the `GetAvatarUrls` function in the `DCLAvatarTexture` class? https://github.com/decentraland/unity-renderer/blob/dev/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Textures/DCLAvatarTexture.cs#L90

sandrade-dcl: As I can see, it is simply doing a request using this:

```
Environment.i.platform.serviceProviders.catalyst.lambdasUrl + "/profiles?id=" + model.userId
```

menduz: All of them are things that I'd like to prevent. The information flows should be way easier than the proposed solution.

The best scenario is adding a `/lambdas/profile-pictures/:address` url to make the whole integration completely stateless.

All this state is damaging for the maintenance of the project(s) and increment the surface area of the responsibilities of kernel and matrix, and we may want to reduce that as much as possible.

lorux0: We do need the profile picture url for being able to display it on the member list, we will need to solve it on renderer side anyways so i dont see an advantage there. On the other hand, we dont need profile pictures to be able to display chat messages, so that case is avoidable. However this change would require a refactor on the renderer side, risking the current solution though

packages/shared/friends/sagas.ts
guidota: would need to rework the `matrix-client` API and ensure somehow that the field will always be present, not sure right now if that's possible

packages/shared/friends/sagas.ts
```diff
@@ -1550,38 +1591,62 @@ export async function getChannelMessages(request: GetChannelMessagesPayload) {
...
       sender,
+      senderName: membersById.get(sender),
       recipient: request.channelId
     })
   }
 
   getUnityInstance().AddChatMessages(addChatMessages)
 }
 
+function findMissingMembers(members: [string, string][], ownId: string) {
```
moliva, line 1636: Can we give an alias to the `[string, string]` type here?

packages/shared/friends/sagas.ts
```diff
@@ -1716,12 +1781,18 @@ export function muteChannel(muteChannel: MuteChannelPayload) {
     store.dispatch(unmutePlayers([channelId]))
   }
 
+  let onlineMembers = 0
+  if (channel.userIds) {
+    const userStatuses = client.getUserStatuses(...channel.userIds)
+    onlineMembers += [...userStatuses.values()].filter((status) => status.presence === PresenceType.ONLINE).length
+  }
+
```
moliva, line 1789: This is duplicating the logic in lines 399 and forth. Let's extract an util function for counting online users.

packages/shared/friends/sagas.ts
```diff
@@ -1753,73 +1823,128 @@ export function getChannelInfo(request: GetChannelInfoPayload) {
...
+    const channel = client.getChannel(channelId)
+    if (!channel) continue
+
+    const muted = profile?.muted?.includes(channelId) ?? false
+
+    let onlineMembers = 0
+    if (channel.userIds) {
+      const userStatuses = client.getUserStatuses(...channel.userIds)
+      onlineMembers += [...userStatuses.values()].filter((status) => status.presence === PresenceType.ONLINE).length
+    }
```
moliva, line 1840: Same as above regarding the logic dup.

packages/shared/friends/sagas.ts
```diff
@@ -1550,38 +1591,62 @@ export async function getChannelMessages(request: GetChannelMessagesPayload) {
...
   const messages = cursorMessage.getMessages()
   if (messageId !== undefined) {
     // we remove the messages they already have.
     const index = messages.map((messages) => messages.id).indexOf(messageId)
     messages.splice(index)
   }
+  const ownId = client.getUserId()
+
+  // deduplicate sender IDs
+  const senderIds = [...new Set(messages.map((message) => message.sender))]
```
menduz, line 1606: Array.from

packages/shared/friends/sagas.ts
```diff
@@ -1753,73 +1823,128 @@ export function getChannelInfo(request: GetChannelInfoPayload) {
...
-  store.dispatch(addedProfilesToCatalog(filteredProfiles.map((profile) => profile.data)))
+  getUnityInstance().UpdateChannelMembers(channelMembersPayload)
+}
 
-  getUnityInstance().UpdateChannelMembers(channelMembers)
+/*
+ * Checks which members are present in the profile catalog and sends partial profiles for missing users
+ * @param members is an array of [member ID, name]
+ */
+function sendMissingProfiles(members: [string, string][], ownId: string) {
```
menduz, line 1907: Can this be more specific? for maintainability purposes, code is written once and read one thousand times

```suggestion
function sendMissingProfiles(members: {address: string, ownId: string}[], ownId: string) {
```

packages/shared/friends/sagas.ts
```diff
@@ -1753,73 +1823,128 @@ export function getChannelInfo(request: GetChannelInfoPayload) {
...
+  // find missing users
+  const missingUsers = findMissingMembers(members, ownId)
+
+  if (missingUsers.length > 0) {
+    const missingProfiles = getMissingProfiles(missingUsers)
+    getUnityInstance().AddUserProfilesToCatalog({ users: missingProfiles })
```
menduz, line 1913: Is this still absolutely necessary? I'm afraid that it will be extremely difficult to remove and I wouldn't like to risk to maintain this catalog bound forever to the chat

packages/shared/friends/sagas.ts
guidota: It's done in this PR: https://github.com/decentraland/kernel/pull/614/files#diff-bcea93a768b74b3a44000978a9f8e8f2b257395933ac042653b8153f0a1c751fR1979

packages/shared/friends/sagas.ts
guidota: It should be necessary until we implement the resolution on the Explorer side, which is the tech debt that has already been discussed

packages/shared/friends/sagas.ts
```diff
@@ -387,16 +392,22 @@ function* configureMatrixClient(action: SetMatrixClient) {
...
       case 'join':
-        if (conversation.name === 'Empty room' || !conversation.name) {
+        if (!conversation.name || conversation.name?.startsWith('Empty room')) {
           break
         }
 
+        let onlineMembers = 0
+        if (conversation.userIds) {
+          const userStatuses = client.getUserStatuses(...conversation.userIds)
+          onlineMembers += [...userStatuses.values()].filter((status) => status.presence === PresenceType.ONLINE).length
```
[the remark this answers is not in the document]
Julieta11: haha not at all! we promise to improve it in this [pr](https://github.com/decentraland/kernel/pull/614) 🏃🏼‍♀️

menduz (approved):

menduz (approved):

PR #592, opened 2022-10-11
trigger emote when a wearable has changed

gonpombo8:
# What? 
...

# Why? 
...

### DIFF
--- a/packages/ui/avatar/avatarSystem.ts
+++ b/packages/ui/avatar/avatarSystem.ts
@@ -10,6 +10,7 @@ import {
   // Material,
   // Color4
 } from '@dcl/legacy-ecs'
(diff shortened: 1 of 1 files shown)


packages/ui/avatar/avatarSystem.ts
```diff
@@ -108,9 +123,11 @@ export class AvatarEntity extends Entity {
   }
 
   setExpression(id: string, timestamp: number): void {
-    const shape = this.avatarShape
-    shape.expressionTriggerId = id
-    shape.expressionTriggerTimestamp = timestamp
+    setTimeout(() => {
```
menduz, line 126: this would delay _all_ emotes by 1 sec.

Plus, there is not setTimeout available for the scenes once we go full-sandboxed mode

PR #593, opened 2022-10-11
fix: send multiple channels info on GetChannelInfo

guidota:
# What? 
...

# Why? 
...


packages/shared/friends/sagas.ts
```diff
@@ -1384,10 +1384,14 @@ function* handleJoinOrCreateChannel(action: JoinOrCreateChannel) {
...
-    const { created, conversation } = yield apply(client, client.getOrCreateChannel, [channelId, []])
+    const { created, conversation }: { created: boolean; conversation: Conversation } = yield apply(
+      client,
+      client.getOrCreateChannel,
+      [channelId, []]
+    )
 
     const channel: ChannelInfoPayload = {
-      name: action.payload.channelId,
+      name: conversation.name ?? action.payload.channelId,
```
moliva, line 1394: Is this second branch acceptable or should we always expect the name to be present?

PR #609, opened 2022-10-17
refactor: change tests to use saga directly 

lauti7:
# What? 
This PR refactors some friends' tests to use the saga directly and react to actions instead of exporting a single function to be unit tested. 

# Why? 
Use redux-saga for tests


menduz (approved):
Epic

menduz:
@lauti7 ouch, .only

PR #614, opened 2022-10-18
fix: update channel info when members change

guidota:
# What? 
Bump to `matrix-client` to version `1.18.0 `

Listen for channel members updates and notify the updates to explorer

Send joined members count on channel leave action

Send online members count on get joined channels action

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -319,12 +319,12 @@ function* configureMatrixClient(action: SetMatrixClient) {
         await ensureFriendProfile(senderUserId)
       }
 
-      addNewChatMessage(chatMessage)
-
-      if (message.sender === ownId) {
-        // ignore messages sent by the local user
+      if (message.sender === ownId && !isChannelType) {
+        // ignore messages sent to private chats by the local user
         return
       }
+
+      addNewChatMessage(chatMessage)
 
       if (isChannelType) {
         const muted = profile?.muted ?? []
@@ -998,7 +998,12 @@ function* handleSendChannelMessage(action: SendChannelMessage) {
   try {
     const conversation: Conversation | undefined = yield apply(client, client.getChannel, [channelId])
     if (conversation) {
-      yield apply(client, client.sendMessageTo, [conversation.id, message])
+      const messageId = yield apply(client, client.sendMessageTo, [conversation.id, message.body])
+
+      if (messageId) {
+        message.messageId = messageId
+      }
+      getUnityInstance().AddMessageToChatWindow(message)
     }
   } catch (e: any) {
     logger.error(e)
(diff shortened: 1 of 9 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -389,18 +389,18 @@ function* configureMatrixClient(action: SetMatrixClient) {
     handleIncomingFriendshipUpdateStatus(FriendshipAction.REJECTED, socialId)
   )
 
+  client.onChannelMembers((conversation) => {
+    updateChannelInfo(conversation, client)
```
Julieta11, line 393: Shouldn't we also send the update about the members list with the `UpdateChannelMembers` message? I think that's the message that will solve the blank member list and toggle issue.

Julieta11 (approved):
🙌🏼 🏃🏼‍♀️! I just left a quick question comment

packages/shared/friends/sagas.ts
```diff
@@ -389,18 +389,26 @@ function* configureMatrixClient(action: SetMatrixClient) {
     handleIncomingFriendshipUpdateStatus(FriendshipAction.REJECTED, socialId)
   )
 
+  client.onChannelMembers((conversation, members) => {
+    updateChannelInfo(conversation, client)
```
Julieta11, line 393: I think before we do anything we can make use of this validation here: 
```
if (!conversation.name || conversation.name?.startsWith('Empty room')) {
          break
}
```

packages/shared/friends/sagas.ts
```diff
@@ -389,18 +389,28 @@ function* configureMatrixClient(action: SetMatrixClient) {
...
 
+  client.onChannelMembers((conversation, members) => {
+    if (conversation.name && !conversation.name.startsWith('Empty room')) {
+      updateChannelInfo(conversation, client)
+    }
+
+    const userStatuses = client.getUserStatuses(...members.map((member) => member.userId))
+    const membersPayload = members
+      .filter((member) => userStatuses.get(member.userId)?.presence === PresenceType.ONLINE)
+      .map((member) => ({ ...member, isOnline: true }))
```
moliva, line 400: I'm wondering if there's a way we can reuse the logic from the `getOnlineMembers` or some intermediate function.

menduz (dismissed):

packages/shared/friends/sagas.ts
```diff
@@ -389,18 +389,39 @@ function* configureMatrixClient(action: SetMatrixClient) {
...
+
+    const channelMembers: ChannelMember[] = members
+      .filter((member) => onlineMemberIds.includes(member.userId))
+      .map((member) => ({
+        userId: getUserIdFromMatrix(member.userId),
+        name: member.name,
+        isOnline: true
+      }))
+
+    const update: UpdateChannelMembersPayload = { channelId: conversation.id, members: channelMembers }
```
menduz, line 411: for future features, I'd recommend to put this logic inside dcl-matrix-client

menduz (approved):

PR #626, opened 2022-10-24
Check if a realm is blacklisted before connecting when changing realm

hugoArregui:
Also remove some debug logs

### DIFF
--- a/packages/shared/dao/index.ts
+++ b/packages/shared/dao/index.ts
@@ -18,6 +18,7 @@ import {
 } from 'shared/realm/resolver'
 import { AboutResponse } from '@dcl/protocol/out-ts/decentraland/bff/http_endpoints.gen'
 import { OFFLINE_REALM } from 'shared/realm/types'
+import { getDisabledCatalystConfig } from 'shared/meta/selectors'
 
 async function fetchCatalystNodes(endpoint: string | undefined): Promise<CatalystNode[]> {
   if (endpoint) {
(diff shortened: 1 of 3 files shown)


packages/shared/dao/index.ts
```diff
@@ -198,11 +195,18 @@ export async function resolveRealmConfigFromString(realmString: string) {
...
+  const denylistedCatalysts: string[] = getDisabledCatalystConfig(store.getState()) ?? []
   const realmConfig = await resolveRealmConfigFromString(realmString)
 
   if (!realmConfig) {
     throw new Error(`The realm ${realmString} isn't available right now.`)
   }
+  const catalystURL = new URL(realmConfig.baseUrl)
+
+  if (denylistedCatalysts.find((denied) => new URL(denied).host === catalystURL.host)) {
+    commsLogger.log(`The realm is blacklisted.`)
```
menduz, line 207: Denylisted

packages/shared/dao/index.ts
```diff
@@ -198,11 +195,18 @@ export async function resolveRealmConfigFromString(realmString: string) {
...
+    return
```
menduz, line 208: Throw?

hugoArregui: when I throw it's catch by the debugger, I though this is better

menduz: The throw appears as a message in the chat, the log in the other hand only appears in the console

packages/shared/dao/index.ts
menduz, line 207:
```suggestion
    commsLogger.log(`The realm is denylisted.`)
```

PR #649, opened 2022-10-28
fix: get DMs by FriendshipStatus instead of Membership

Julieta11:
# What? 
Bump `dcl-social-client` to version `1.20.1` in oder to get DMs filtered by FriendshipStatus instead of room Membership

### DIFF
--- a/packages/shared/friends/selectors.ts
+++ b/packages/shared/friends/selectors.ts
@@ -5,18 +5,20 @@ import { getUserIdFromMatrix } from './utils'
 
 export const getSocialClient = (store: RootFriendsState) => store.friends.client
 
+/**
+ * Get all conversations `ConversationType.CHANNEL` that the user has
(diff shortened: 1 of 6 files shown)


packages/shared/friends/selectors.ts
```diff
@@ -36,6 +38,10 @@ export const getConversations = (
     }))
 }
 
+/**
+ * Get all current conversations with messages the user has including DMs, channels, etc
+ * @return `conversation` & `unreadMessages` boolean that indicates whether the conversation has unread messages.
+ */
 export const getAllConversationsWithMessages = (
```
Julieta11, line 45: We are not referencing this anywhere. Should we keep it? What do you think reviewer? 😛

gonpombo8: 🔪 kill it

PR #658, opened 2022-11-01
fix track event moveToparcel

gonpombo8:
# What? 
Fix moveToParcel track event reporting 0,0 when you start
...

# Why? 

Kernel is reporting that we started on Genesis Plaza, and then we teleport out into the tile we actually had in the url, messing with our metrics
...

image

### DIFF
--- a/packages/shared/scene-loader/sagas.ts
+++ b/packages/shared/scene-loader/sagas.ts
@@ -180,14 +180,16 @@ function* rendererPositionSettler() {
       getPositionSpawnPointAndScene
     )
 
-    // and then settle the position
-    if (!isSettled) {
-      // Then set the parcel position for the scene loader
-      receivePositionReport(spawnPointAndScene.spawnPoint.position)
+    if (!!spawnPointAndScene.sceneId) {
+      // and then settle the position
+      if (!isSettled) {
+        console.log('Move to parcel, Settled,', spawnPointAndScene, isSettled)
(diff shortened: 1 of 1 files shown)


packages/shared/scene-loader/sagas.ts
```diff
@@ -180,14 +180,16 @@ function* rendererPositionSettler() {
...
     )
 
-    // and then settle the position
-    if (!isSettled) {
-      // Then set the parcel position for the scene loader
-      receivePositionReport(spawnPointAndScene.spawnPoint.position)
+    if (!!spawnPointAndScene.sceneId) {
+      // and then settle the position
+      if (!isSettled) {
+        console.log('Move to parcel, Settled,', spawnPointAndScene, isSettled)
```
menduz, line 186:
```suggestion
```

packages/shared/scene-loader/sagas.ts
```diff
@@ -180,14 +180,16 @@ function* rendererPositionSettler() {
...
+        // Then set the parcel position for the scene loader
+        receivePositionReport(spawnPointAndScene.spawnPoint.position)
+      }
+      // then update the position in the engine
+      getUnityInstance().Teleport(spawnPointAndScene.spawnPoint)
     }
```
menduz, line 192:
```suggestion
    }
    // then update the position in the engine
    getUnityInstance().Teleport(spawnPointAndScene.spawnPoint)
```

PR #664, opened 2022-11-03
fix: hide fps panel when SCENE_DEBUG_MODE is off

dalkia:
# What? 

Stops the FPS panel from showing in Preview or Editor mode if the `SCENE_DEBUG_MODE` param is off.

The compromise is that `SHOW_FPS_COUNTER` param will not work in Preview or Editor mode. But there are many ways to toggle the debug panel (/showfps, click on the [<>] icon, use the `SCENE_DEBUG_MODE` param). 

We may consider a new PR with the cleaning of the DEBUG params, since it requires a little bit of organizing.

# Why? 

This was required on this chat. https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1667417735159079


cazala (approved):

packages/config/index.ts
```diff
@@ -122,7 +122,7 @@ export const RESET_TUTORIAL = location.search.includes('RESET_TUTORIAL')
 
 export const ENGINE_DEBUG_PANEL = location.search.includes('ENGINE_DEBUG_PANEL')
 export const SCENE_DEBUG_PANEL = location.search.includes('SCENE_DEBUG_PANEL') && !ENGINE_DEBUG_PANEL
-export const SHOW_FPS_COUNTER = location.search.includes('SHOW_FPS_COUNTER') || DEBUG
+export const SHOW_FPS_COUNTER = (location.search.includes('SHOW_FPS_COUNTER') || DEBUG) && WORLD_EXPLORER
```
menduz, line 125: I think this may be a better condition, reverting the changes in `packages/unity-interface/dcl.ts` to respect the single responsibility of `SHOW_FPS_COUNTER`

```suggestion
export const SHOW_FPS_COUNTER = location.search.includes('SHOW_FPS_COUNTER') || SCENE_DEBUG_PANEL
```

dalkia: Is there some way to check which is the most used `param`?

dalkia: WDYT if we separate the EDITOR and PREVIEW booleans from the Debug section? With the latest configuration I pushed, we get the fps panel with DEBUG_MODE on; and it's not going to be dependant on the value of SCENE_DEBUG_PANEL.

packages/config/index.ts
```diff
@@ -105,15 +105,15 @@ export const TRACE_RENDERER = ensureSingleString(qs.get('TRACE_RENDERER'))
 
 export const LOS = ensureSingleString(qs.get('LOS'))
 
-export const DEBUG = location.search.includes('DEBUG_MODE') || !!(globalThis as any).mocha || PREVIEW || EDITOR
+export const DEBUG = location.search.includes('DEBUG_MODE') || !!(globalThis as any).mocha
```
menduz, line 108: `DEBUG` is required to be true in preview mode

PR #670, opened 2022-11-08
Feat: Add social service url and feature flag

### LINKED ISSUE #650: Setup a feature flag to point users to new social service

jmoguilevsky:
This PR adds the social server feature, it also adds a feature flag to roll out of the usage of synapse. This PR also allows for usage of a different social server

Closes #650 
...

# Why? 
...

### DIFF
--- a/packages/config/index.ts
+++ b/packages/config/index.ts
@@ -218,6 +218,7 @@ function getDefaultAssetBundlesBaseUrl(network: ETHEREUM_NETWORK): string {
 
 export function getAvatarTextureAPIBaseUrl(network: ETHEREUM_NETWORK): string {
   const tld = network === ETHEREUM_NETWORK.MAINNET ? 'org' : 'zone'
+  // TODO!: Change this to point to social once the rollout is complete
   return `https://synapse.decentraland.${tld}/profile-pictures/`
 }
 
@@ -229,9 +230,13 @@ export function getServerConfigurations(network: ETHEREUM_NETWORK) {
   const questsUrl =
     ensureSingleString(qs.get('QUESTS_SERVER_URL')) ?? `https://quests-api.decentraland.${network ? 'org' : 'io'}`
 
+  const socialServerUrl =
+    ensureSingleString(qs.get('SOCIAL_SERVER_URL')) ?? `https://social.decentraland.${tld}`
+
   return {
     explorerConfiguration: `${metaConfigBaseUrl}?t=${new Date().getTime()}`,
-    questsUrl
+    questsUrl,
+    socialServerUrl
   }
 }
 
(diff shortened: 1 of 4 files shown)


packages/shared/meta/sagas.ts
```diff
@@ -160,9 +163,9 @@ async function fetchMetaConfiguration(network: ETHEREUM_NETWORK) {
         denied: [],
         contentWhitelist: []
       },
-      bannedUsers: {},
```
menduz, line 163: why is this being removed?

packages/shared/meta/sagas.ts
```diff
@@ -138,12 +138,15 @@ async function fetchFeatureFlagsAndVariants(network: ETHEREUM_NETWORK): Promise<
...
+async function fetchMetaConfiguration(network: ETHEREUM_NETWORK): Promise<Partial<MetaConfiguration>> {
+  const serverConfiguration = getServerConfigurations(network)
+  const explorerConfigurationEndpoint = serverConfiguration.explorerConfiguration
+  const socialServerUrl = serverConfiguration.socialServerUrl
+
   try {
     const response = await fetch(explorerConfigurationEndpoint)
     if (response.ok) {
-      return response.json()
+      return { ...response.json(), socialServerUrl }
```
menduz, line 149: this can be prevented in the selector. if the response is not OK then it will skip the configuration override entirely

packages/config/index.ts
```diff
@@ -229,9 +230,13 @@ export function getServerConfigurations(network: ETHEREUM_NETWORK) {
   const questsUrl =
     ensureSingleString(qs.get('QUESTS_SERVER_URL')) ?? `https://quests-api.decentraland.${network ? 'org' : 'io'}`
 
+  const socialServerUrl =
+    ensureSingleString(qs.get('SOCIAL_SERVER_URL')) ?? `https://social.decentraland.${tld}`
```
menduz, line 234: please move this into the selector

menduz (changes requested):
Although the code may work, it is best to not have any extra state than needed. I recommend to move all the logic into the selector `getSynapseUrl` directly, skipping any storage of extra fields in the configuration

packages/config/index.ts
jmoguilevsky: moved into selector

packages/shared/meta/sagas.ts
jmoguilevsky: moved into selector

packages/config/index.ts
```diff
@@ -79,7 +79,7 @@ function ensureQueryStringUrl(value: string | null): string | null {
   if (typeof value === 'string') return addHttpsIfNoProtocolIsSet(value)
   return addHttpsIfNoProtocolIsSet(value[0])
 }
-function ensureSingleString(value: string | string[] | null): string | null {
+export function ensureSingleString(value: string | string[] | null): string | null {
```
menduz, line 82:
```suggestion
function ensureSingleString(value: string | string[] | null): string | null {
```

menduz (approved):

PR #679, opened 2022-11-09
feat: bump `dcl-social-client` to version `1.20.3`

Julieta11:
# What? 
We currently have concurrent update errors for different users when there are multiple rooms trying to set up `m.direct` on the initial sync.

So, to avoid the concurrent update error, we decided to set the user's `m.direct` events all at once with a single request, which allows us to lighten the number of updates and thus reduce the concurrent update error.

### DIFF
--- a/packages/shared/friends/selectors.ts
+++ b/packages/shared/friends/selectors.ts
@@ -39,29 +39,6 @@ export const getConversations = (
 }
 
 /**
- * Get all current conversations with messages the user has including DMs, channels, etc
- * @return `conversation` & `unreadMessages` boolean that indicates whether the conversation has unread messages.
- */
-export const getAllConversationsWithMessages = (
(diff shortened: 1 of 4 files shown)


packages/shared/friends/selectors.ts
```diff
@@ -38,29 +38,6 @@ export const getConversations = (
     }))
 }
 
-/**
- * Get all current conversations with messages the user has including DMs, channels, etc
- * @return `conversation` & `unreadMessages` boolean that indicates whether the conversation has unread messages.
- */
-export const getAllConversationsWithMessages = (
```
Julieta11, line 45: I'm killing this one: https://github.com/decentraland/kernel/pull/649#discussion_r1008429971

menduz (approved):

PR #680, opened 2022-11-09
feat: return joined members when presence is disabled

moliva:
# What? 
When the feature flag for presence is disabled, show the joined members of a channel instead of the online ones (as this information will not be available).

# Why? 
This is used as a mitigation tool to provide the best possible experience to the end user when the backend service cannot provide this feature.

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -397,16 +397,16 @@ function* configureMatrixClient(action: SetMatrixClient) {
       updateChannelInfo(conversation, client)
     }
 
-    // we only notify members who are online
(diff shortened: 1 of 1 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -1997,6 +2014,16 @@ function isAllowedToCreate() {
   }
 }
 
+function getOnlineOrJoinedMembers(userIds: string[], client: SocialAPI): string[] {
```
guidota, line 2017: is this function duplicated?

guidota (approved):
I think there is a duplicated function, but it looks good to me besides that.

menduz (approved):

PR #692, opened 2022-11-14
fix: reach pagination limit

### LINKED ISSUE #673: Not all connected members appear in the member list of a channel
Sometimes there is people currently talking in a specific chat channel which is not displayed in the member list.

Repro steps:
1. Join a chat channel through the channel browser
2. Open the member list of the channel
3. Receive message in the channel from different people
4. Sometimes the person is not displayed in the member list, being online at the moment

Julieta11:
Closes #673 

# What? 
When the FF `matrix_presence_disabled` is deactivated, members list pagination does not work as expected, as it first paginates and then checks the presence status of members, leading to possible incorrect pagination.

i.e., the channel has 100 members joined, paging goes from 10 to 10 and we have 5 online members on page 1 and other 4 online members on page 2. We were retrieving only 5 members on the first request when we should be retrieving all 9 online members on the first request.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "@dcl/kernel-interface": "^2.0.0-20210922153939.commit-017905d",
         "@dcl/legacy-ecs": "^6.11.8",
         "@dcl/protocol": "^1.0.0-3373991894.commit-8aa3a49",
-        "@dcl/rpc": "^1.0.4-20221113192916.commit-3ef5187",
+        "@dcl/rpc": "^1.1.1-20221115000939.commit-9a51ad0",
         "@dcl/scene-runtime": "^1.0.0-20221102005705.commit-05d463c",
         "@dcl/schemas": "^5.21.0",
         "@dcl/urn-resolver": "^2.0.2",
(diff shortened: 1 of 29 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -1928,36 +1928,37 @@ export async function getChannelMembers(request: GetChannelMembersPayload) {
...
 
-  const ownId = client.getUserId()
-
-  // update catalog with missing users, by using default profiles with name and image url
-  sendMissingProfiles(members, ownId)
-
   // we only notify members who are online if presence is enabled, else every joined member
-  const memberIds = members.map((member) => member.userId)
+  const memberIds = allMembers.map((member) => member.userId)
   const onlineOrJoinedMemberIds = getOnlineOrJoinedMembers(memberIds, client)
```
Julieta11, line 1945: This is not the best approach especially when there are many members joined, but I thought of it that way because at the moment the explorer team sets the skip value based on how many members they have loaded. We wouldn't be able to know what our last page was, at least we changed the payloads so that we send each other that information. Any thoughts reviewers?

Issue #689, opened 2022-11-13
Users joined to the #mvmf channel are noticing huge lags


sandrade-dcl:
For reasons as yet unknown, all users joined to the #mvmf channel are noticing huge lags. Once they leave this specific channel, the app stops experiencing it and returns to normal.

We have decided to temporally disable the channels feature until find a solution.

Meanwhile, the users can continue chatting with people through nearby chat and private messages.

### Steps to reproduce it
After disabling the channels FF, we're not able to reproduce it from https://play.decentraland.org/?ENABLE_matrix_channels_enabled

Issue #688, opened 2022-11-13
App freezes after joining #mvmf channel


sandrade-dcl:
The app keeps completely freezed after joining the #mvmf channel. It seems something related to the number of joined members of the channel (#mvmf has 12K+ joined members) because it is not happening with the rest of the channels (few joined members).

![image.png](https://images.zenhubusercontent.com/5eb2b8ca65c88f6e391ef16e/8dd0a4ab-f935-4e4f-8019-eff791c15f1e)

### Steps to repdroduce it (it only happens from build, not from Unity editor):
1. Be sure you're not joined into the #mvmf channel.
2. Go to CHANNEL BROWSER.
3. Click on the JOIN button of the #mvmf channel.
4. Notice how, after some seconds, the app freezes. 
5. If you refresh you will see that the join was done successfully.

### Notes
After debugging the problem from Unity editor, we cannot reproduce the issue from the editor neither we identified any possible cause. It worth it to debug from the kernel side and see what is happening after the join is done.

PR #701, opened 2022-11-14
fix: do not send channel member update on every member

guidota:
Solves #689 
Probably solves #688 

# What? 
Remove the listener where every member change is sent to Explorer.

# Why? 
In cases like joining a big channel is not well handled, sending N updates, where N is the number of members of that channel. 
The changes now are only sent when requested by Explorer.


Julieta11 (approved):
Great catch, Guido! 🏄🏼

leanmendoza (approved):

PR #704, opened 2022-11-14
feat: jump in home from realm 

dalkia:
# What? 

Adds a Jump Home command with a change realm

# Why? 

Necessary for worlds feature, where you need to be able to jump to another realm and then go home

### DIFF
--- a/packages/shared/dao/index.ts
+++ b/packages/shared/dao/index.ts
@@ -239,6 +239,7 @@ export async function changeRealm(realmString: string, forceChange: boolean = fa
   return
 }
 
+// TODO: unify this function with the one implementing the realm selection algorithm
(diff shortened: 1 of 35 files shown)


dalkia:
We should consider adding this wait function before any `goto `command; as we may be in a world and need a realm change

packages/shared/dao/index.ts
```diff
@@ -239,4 +239,33 @@ export async function changeRealm(realmString: string, forceChange: boolean = fa
   return
 }
 
+export async function changeToMostPopulatedRealm(): Promise<void> {
```
menduz, line 242:
```suggestion
// TODO: unify this function with the one implementing the realm selection algorithm
export async function changeToMostPopulatedRealm(): Promise<void> {
```

menduz (approved):

PR #710, opened 2022-11-15
feat: send `InitializeChat` message when a guest logs in

### LINKED ISSUE #717: Join channels by name is not working
## Issue
When we try to join a channel clicking a link posted in our chats, Kernel is returning an unknown error. This is because, in this specific case, Unity is sending to Kernel the message `JoinOrCreateChannel` with the NAME of the channel, and this method only works with the ID of the channel.

### Suggested solution
The message `JoinOrCreateChannel` should support both names and ids.

### Steps to reproduce
1. Ensure your wallet have permissions to create channels (it is configured in the `users_allowed_to_create_channels` feature flag's variant).
2. Ensure you are not joined to the `#global` channel.
3. Write in any chat the text `#global` to create a channel link.
4. Click on the channel link.
5. Notice the Join Channel modal appears.
6. Click on the JOIN button.
7. Observe the issue.

### Notes
If we execute the join through command line (`/join global`), it works perfectly.

Julieta11:
When a guest user logs in, the explorer team has problems with the chat because it works fine (`~nearby`), but it is flagged as uninitialized. 

Matrix doesn't support guest users, so we skip private messaging initialization as long as it's a guest, but we figured that we can send the message `InitializeChat` without causing any trouble. In the meantime, we'll keep working on the protocol to handle these cases more appropriately.

----

We have added logic to support the ability to join a channel using a deep link and a bump of `dcl-social-client` to version `1.20.4` to fix the issue: _'join channels by name is not working 717'_ as well.

Closes #717


packages/shared/friends/sagas.ts
```diff
@@ -176,7 +176,10 @@ function* initializeFriendsSaga() {
     const isGuest = yield select(getIsGuestLogin)
 
     // guests must not use the friends & private messaging features
-    if (isGuest) return
+    if (isGuest) {
+      getUnityInstance().InitializeChat({ totalUnseenMessages: 0 })
```
moliva, line 180: It might be better for present usage as well as the future to have some kind of flag or kind here, i.e. `isGuest: true | false` or `kind: `User.WALLET | User.Guest`, to have a better understanding of how to handle this on the renderer side.

lorux0: it is useful, but not needed at the moment though

menduz (approved):

PR #712, opened 2022-11-15
Revert "fix: rpc race condition"

menduz:
Reverts decentraland/kernel#702

It seems like the released version of the renderer doesn't respond the binary messages over web sockets but the editor version does.


kuruk-mm (approved):

kuruk-mm:
Force push to verify the commit

PR #728, opened 2022-11-22
feat: add `analyticsKernelService` with renderer-protocol

kuruk-mm:
# What? 
Add analyticsKernelService using renderer-protocol

# Why? 
...


packages/unity-interface/initializer.ts
```diff
@@ -101,7 +101,6 @@ async function loadInjectedUnityDelegate(
   const transport = webTransport({ wasmModule: originalUnity.Module }, (globalThis as any).DCL)
 
   await engineStartedFuture
-  await browserInterface.startedFuture
```
kuruk-mm, line 104: this was preventing the RPC from starting...

this is already being awaited here: https://github.com/decentraland/kernel/blob/e9c174b4e2ae20fba7309a3703778c434da41b65/packages/unity-interface/dcl.ts#L48

pbosio (approved):

packages/unity-interface/initializer.ts
menduz: Does it still work in desktop? I remember this was a bugfix

PR #731, opened 2022-11-22
feat: Add protocol in PerformanceInfo

agusaldasoro:
Add comms protocol tag in metrics for ping


packages/shared/comms/sagas.ts
```diff
@@ -202,6 +207,7 @@ function* handleConnectToComms(action: ConnectToCommsAction) {
 
     let adapter: RoomConnection | undefined = undefined
 
+    overrideCommsProtocol(protocol)
```
menduz, line 210:
```suggestion
    // TODO: move this to a saga
    overrideCommsProtocol(protocol)
```

menduz (approved):

PR #734, opened 2022-11-28
fix: get `lastStatusOfFriends` at `refreshFriends`

Julieta11:
We initialize lastStatusOfFriends as an empty map and it makes sense when we start the app. But then while the app is running we trigger the refreshFriends saga ([check](https://github.com/decentraland/kernel/blob/c2611bc29c6ef34930b9af5d9dc5692b93a893d9/packages/shared/friends/sagas.ts#L1212-L1215)) several times. To dedupe the number of messages sent we need to check if we have current states stored and keep them if we do or initialize the empty map if we don't.


moliva (approved):

pravusjif (approved):
tested without problems

PR #740, opened 2022-11-29
feat: new-friend-request

Julieta11:
# What? 
...

# Why? 
...

### DIFF
--- a/packages/shared/friends/sagas.ts
+++ b/packages/shared/friends/sagas.ts
@@ -16,7 +16,7 @@ import {
   GetOrCreateConversationResponse
 } from 'dcl-social-client'
 
-import { DEBUG_KERNEL_LOG, CHANNEL_TO_JOIN_CONFIG_URL } from 'config'
+import { DEBUG_KERNEL_LOG, ethereumConfigurations, CHANNEL_TO_JOIN_CONFIG_URL } from 'config'
 import { deepEqual } from 'atomicHelpers/deepEqual'
 
 import defaultLogger, { createLogger, createDummyLogger } from 'shared/logger'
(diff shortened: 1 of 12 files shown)


packages/shared/friends/sagas.ts
```diff
@@ -537,14 +554,18 @@ function* refreshFriends() {
...
+        friendRequestId: `${getUserIdFromMatrix(request.from)}_${getUserIdFromMatrix(ownId)}`, // {from_requester}_{to_requested} -> 'otherUserId_ownId',
         createdAt: request.createdAt,
-        userId: getUserIdFromMatrix(request.from)
+        userId: getUserIdFromMatrix(request.from),
+        message: request.message
       })
     )
     const requestedToIds = toFriendRequests.map(
       (request): FriendRequest => ({
+        friendRequestId: `${getUserIdFromMatrix(ownId)}_${getUserIdFromMatrix(request.to)}`, // {from_requester}_{to_requested} -> 'ownId_otherUserId',
```
moliva, line 565: I'm thinking it might be better to have the same criterion as for the friendship channel, i.e. `ownId < otherUserId ? ownId_otherUserId : otherUserId_ownId`. WDYT?
Also let's extract this pattern into a function to keep the same format everywhere we need to generate the request id.

packages/shared/friends/sagas.ts
```diff
@@ -1164,6 +1198,18 @@ function* handleUpdateFriendship({ payload, meta }: UpdateFriendship) {
           totalReceivedRequests: updateTotalFriendRequestsPayload.totalReceivedRequests + 1
         }
 
+        // We only send this message in the new flow
+        if (newFriendRequestFlow) {
```
moliva, line 1202: Let's add a `TODO` for all these conditionals to remember removing this in the near future.

PR #744, opened 2022-11-30
feat: handle WsPeerKicked message

marianogoldman:


gonpombo8 (approved):

packages/shared/comms/adapters/WebSocketAdapter.ts
```diff
@@ -99,6 +100,11 @@ export class WebSocketAdapter implements MinimumCommunicationsAdapter {
             )
             break
           }
+          case 'peerKicked': {
+            const { peerKicked } = message
+            notifyStatusThroughChat(peerKicked.reason)
+            throw new Error(message.peerKicked.reason)
```
menduz, line 106: This should trigger a disconnect({ reason }) event

PR #745, opened 2022-11-30
feat: Teleport worlds commands

dalkia:
Adds features required in [#3547](https://app.zenhub.com/workspaces/explorer-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/3547)

# What? 

Adds a realm check before any goTo command to go back to a DAO realms if necessary
Adds a `/world name` command to go directly to a world. No ENS required
Function `GoToMostPopulatedRealm()` was returning the least populated realm. Changed it back

# Why?

### DIFF
--- a/packages/shared/chat/sagas.ts
+++ b/packages/shared/chat/sagas.ts
@@ -270,7 +270,7 @@ function initChatCommands() {
   })
 
   addChatCommand('world', 'Goes to a world', (message) => {
-    const worldString = `${message.trim()}.dcl.eth`
+    const worldString = `${message.trim()}`
     const response = ''
 
     changeRealm(worldString).catch((e) => {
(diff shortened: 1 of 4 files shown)


packages/shared/chat/sagas.ts
```diff
@@ -263,6 +269,24 @@ function initChatCommands() {
     }
   })
 
+  addChatCommand('world', 'Goes to a world', (message) => {
+    const worldString = `${message.trim()}.dcl.eth`
```
menduz, line 273:
```suggestion
    const worldString = `${message.trim()}`
```

menduz (changes requested):
Worlds are realms, and realms need to be resolved by ENS only

PR #746, opened 2022-11-30
feat: cancel friend request

### LINKED ISSUE #709: Cancel friend request
- [ ] Add tests for RPC flow

Julieta11:
# What? 

So far, the PR addresses:
- Cancel friend request - Closes #709


packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
```diff
@@ -109,7 +110,52 @@ export function registerFriendRequestKernelService(port: RpcServerPort<RendererP
...
+            message: {
+              $case: 'reply',
+              reply: {
+                friendRequest: {
+                  friendRequestId: cancelFriend.reply.friendRequest.friendRequestId,
+                  timestamp: cancelFriend.reply.friendRequest.timestamp,
+                  to: cancelFriend.reply.friendRequest.to,
+                  from: cancelFriend.reply.friendRequest.from,
+                  messageBody: cancelFriend.reply.friendRequest.messageBody
+                }
```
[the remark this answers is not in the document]
Julieta11: Definitely!

packages/shared/friends/sagas.ts
```diff
@@ -2153,6 +2158,57 @@ export async function UpdateFriendshipAsPromise(
...
+    const response = await UpdateFriendshipAsPromise(FriendshipAction.CANCELED, userId.toLowerCase(), false)
+
+    if (!response.error) {
+      const sendFriendRequest: CancelFriendRequestReplyOk = {
+        friendRequest: {
+          friendRequestId: request.friendRequestId,
+          timestamp: Date.now(),
+          from: getUserIdFromMatrix(ownId),
+          to: userId,
+          messageBody
```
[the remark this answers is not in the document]
Julieta11: I kind of had the same question when we were designing the RFC and Nico told me they don't have that kind of info in the context in which they're receiving the msg. (check it out [here](https://github.com/decentraland/adr/pull/146#discussion_r1037537280))

packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
```diff
@@ -72,7 +73,32 @@ export function registerFriendRequestKernelService(port: RpcServerPort<RendererP
...
+        let cancelFriendRequestReply: CancelFriendRequestReply = {}
+
+        // Check the response type
+        if (cancelFriend.reply?.friendRequest) {
+          cancelFriendRequestReply = {
+            message: {
+              $case: 'reply',
+              reply: cancelFriend.reply
+            }
+          }
```
moliva, line 89: Wondering whether we can do the same here as with the error function for all replies alike.

moliva (approved):
Leaving a comment for a potential refactor but other than that it's great! 👍🏻

packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
```diff
@@ -114,14 +84,37 @@ export function registerFriendRequestKernelService(port: RpcServerPort<RendererP
...
+}
+
+/**
+ * Build friend requests message to send to renderer.
+ * If the friendRequest object is truthy, the function returns the result of calling `wrapReply` on the `friendRequest` object.
+ * If the friendRequest object is falsy, the function returns the result of calling `buildErrorResponse` with the `error` value.
+ * @param friendRequest - it can represent any kind of FriendRequestReplyOk type.
+ * @param error - an int representing an error code.
+ */
+function buildResponse<T>(friendRequest: T, error?: FriendshipErrorCode) {
```
moliva, line 114: From the use this is given, why not just sending the object with both the reply and error instead (using a single parameter)?

packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
Julieta11: Totally! Take a look at what I have managed to carry out (https://github.com/decentraland/kernel/pull/746/commits/bc24a1b692c22c112b70d1763634234baf234f98)

packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
```diff
@@ -48,27 +37,19 @@ export function registerFriendRequestKernelService(port: RpcServerPort<RendererP
         // Handle send friend request
         const sendFriendRequest = await requestFriendship(req)
 
-        let sendFriendRequestReply: SendFriendRequestReply = {}
-
-        // Check the response type
-        if (sendFriendRequest.reply?.friendRequest) {
```
moliva, line 53: In the new version the `sendFriendRequest.reply` is checked instead of the `sendFriendRequest.reply?.friendRequest`, is this semantically the same?

Julieta11: No, it's not the same. But the older one was unnecessarily checking the friendRequest because if we already have the reply, it means we already have the friendRequest as well

moliva: Great! 👍🏻

packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
Julieta11: I thought about it and I ended up having sth like this because TypeScript was reporting non-existent properties in the generic function:
```ts
type ExtendedType<T, U> = T & { reply?: U; error?: FriendshipErrorCode }

function buildResponse<T, U>(friendRequest: ExtendedType<T, U>) {
  if (friendRequest.reply) {
    return wrapReply(friendRequest.reply)
  } else {
    return buildErrorResponse(friendRequest.error)
  }
}
```

I didn't particularly like the solution tbh. But yes, I think we can definitely go in that direction. Give me a thumbs up and I'll commit that solution.

moliva: Let me check that a sec!

kuruk-mm (approved):

PR #749, opened 2022-12-01
chore: add update-renderer-branch to makefile

kuruk-mm:
add update-renderer-branch to avoid using renderer-branch=main parameter, and iterate faster in the development

use example: `make update-renderer-branch BRANCH=main`

The curl fails with an error with status code != 200

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,14 @@ fetchSceneContents: scripts/fetchSceneContents.js
 update-renderer:  ## Update the renderer
 	npm install @dcl/unity-renderer@latest
 
+# example: make update-renderer-branch BRANCH=main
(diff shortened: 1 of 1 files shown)


menduz:
I rather use the `&renderer-branch=<BRANCH>` query param, this approach sounds very not sustainable. What is the motivation behind it?

kuruk-mm:
> I rather use the `&renderer-branch=<BRANCH>` query param, this approach sounds very not sustainable. What is the motivation behind it?

Faster iteration for the devs, because you don't need to download the branch each time. It's very useful when you have a slow connection, and even when you have a fast one, it takes some seconds to download ~150mb

PR #751, opened 2022-12-01
chore: check for accepting users before joining a realm

hugoArregui:


packages/shared/dao/sagas.ts
```diff
@@ -198,7 +198,8 @@ function* initializeCatalystCandidates() {
 
 export async function checkValidRealm(baseUrl: string): Promise<PingResult | null> {
   const pingResult = await ping(baseUrl + '/about')
-  if (pingResult.status === ServerConnectionStatus.OK) {
+  const acceptingUsers = pingResult.result && (pingResult.result.acceptingUsers ?? true)
```
menduz, line 201: Is this condition checked to return false when acceptingUsers == false?

hugoArregui: not sure to understand the question, sorry

kuruk-mm: I think the doubt comes from:
```ts
(pingResult.result.acceptingUsers ?? true)
```

but

```ts
(false ?? true) = false
```

so it should be ok

menduz (approved):

menduz:
Approving, although this will give no hint about the realm being full. Maybe an exception can be considered to improve the UX

agusaldasoro (approved):

PR #752, opened 2022-12-02
Integrate DCL with IPSME (enacted Grant)

knev:
This PR is part of the grant proposal: Integrating Decentraland with 4 other platforms
https://governance.decentraland.org/proposal/?id=47e1bb80-4139-11ec-be0c-afec86cba5e5

This PR implements IPSME in DCL opening a communication channel between other participants in the eco-system.

The code provides an empty IPSME message handler with DCL devs can use for further development.

*note: Because the latest commit of the kernel does not build properly on my machine, the DCL devs will have to do the merge/rebase.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "@dcl/rpc": "^1.0.1-20220608232056.commit-741737c",
         "@dcl/schemas": "^5.8.1",
         "@dcl/urn-resolver": "^1.4.0",
+        "@ipsme/msgenv-broadcastchannel": "^0.3.2",
         "@redux-saga/simple-saga-monitor": "^1.1.2",
         "@types/redux-logger": "^3.0.9",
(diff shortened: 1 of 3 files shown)


menduz (changes requested):
Hello, thanks for contributing!

This PR seems to do nothing at all, would you expand on how it works?

Also the code of the library you are adding is not public nor used by any other project as per NPMjs. Could you make it open?

knev:
Can you please bring up this conversation on the DAO discord at 
https://discord.com/channels/894658869391933540/938880504164397106

The npm libraries are public, the code repos are not yet. If you install them via npm, you can see all the code.

menduz:
We'll need to review the public repositories before adding any external library. To protect the security of the users. And there must be an implementation plan, we tend to create ADR documents for it https://adr.decentraland.org

PR #760, opened 2022-12-06
feat: send friend request

### LINKED ISSUE #694: Send friend request
- [ ] Add tests for RPC flow

Julieta11:
# What? 

So far, the PR addresses:
- Send friend request - Closes #694


packages/shared/friends/sagas.ts
```diff
@@ -2025,6 +2056,93 @@ export async function getChannelMembers(request: GetChannelMembersPayload) {
   getUnityInstance().UpdateChannelMembers(channelMembersPayload)
 }
 
+export async function requestFriendship(request: SendFriendRequestPayload) {
+  try {
+    let userId = request.userId
+    let found = false
+    const state = store.getState()
+
+    const ownId = store.getState().friends.client?.getUserId()
```
moliva, line 2065: Move this to a selector.

packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
```diff
@@ -59,7 +60,52 @@ export function registerFriendRequestKernelService(port: RpcServerPort<RendererP
...
+              }
+            }
+          }
+        } else {
+          sendFriendRequestReply = {
+            message: {
+              $case: 'error',
+              error: sendFriendRequest.error ?? FriendshipErrorCode.FEC_UNKNOWN
+            }
+          }
```
moliva, line 91: How about a util function that constructs this object with an unknown error by default?

packages/shared/friends/sagas.ts
```diff
@@ -1141,11 +1160,15 @@ function* handleSendPrivateMessage(action: SendPrivateMessage) {
...
-  const { action, userId } = payload
+  const { action, userId, future } = payload
 
   const client: SocialAPI | undefined = yield select(getSocialClient)
 
+  // Get feature flag value
+  const newFriendRequestFlow = isNewFriendRequestEnabled()
+
   if (!client) {
+    yield call(future.resolve, { userId: userId, error: FriendshipErrorCode.FEC_UNKNOWN })
```
moliva, line 1171: Simplify here and below.
```suggestion
    yield call(future.resolve, { userId, error: FriendshipErrorCode.FEC_UNKNOWN })
```

packages/shared/friends/sagas.ts
```diff
@@ -2028,6 +2066,93 @@ export async function getChannelMembers(request: GetChannelMembersPayload) {
...
+        found = true
+      }
+    }
+
+    if (!found) {
+      return { reply: null, error: FriendshipErrorCode.FEC_NON_EXISTING_USER }
+    }
+
+    // Update user data
+    store.dispatch(updateUserData(request.userId.toLowerCase(), getMatrixIdFromUser(request.userId)))
```
moliva, line 2110: Shouldn't it be using the `userId` local variable instead of the request one here?

Julieta11: Actually, we're not requesting the user id, we're only formatting it. Should I change the name to something more telling and less confusing? Do I do it now? What you think?

```
Get the matrixUserId from userId

@param userId — a string with the userId pattern

@example

from: '0x1111ada11111'
to: '@0x1111ada11111:decentraland.org'
```

Julieta11: Never mind— I get it. You're right. Thx

packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
Julieta11: This is the best I can do ([link](https://github.com/decentraland/kernel/pull/760/commits/025d216fdec4fe3b245ab093d92739661d77f049)). If I try to extract this into a more generic func:
```
{
    message: {
      $case: 'error',
      error: error ?? FriendshipErrorCode.FEC_UNKNOWN
    }
}
```
 I get the error:
```
Types of property 'message' are incompatible.
Type '{ $case: string; error: FriendshipErrorCode; }' is not assignable to type '{ $case: "error"; error: FriendshipErrorCode; }'.
 ```

kuruk-mm (approved):
Code looks great!

packages/renderer-protocol/inverseRpc/services/friendRequestService.ts
moliva: It's a bit trickier since the response interfaces are defined for the reply and error formats as a single type. [But here you can find an example of how to unify them](https://github.com/decentraland/kernel/pull/784).

moliva (approved):
LGTM.

PR #763, opened 2022-12-08
Fix docker compose

m3taphysics:
- upgraded to docker-compose version 3
- copy init.sh into the container and apply executable permissions
- add shebang to specify bash interpreter and show errors


menduz (approved):
It may be good to add a pipeline to check this works on github

PR #769, opened 2022-12-12
fix: removing duplicates wearables and limit wearables length

dalkia:
# What? 

Since Unity 2021, there is an issue processing large string for Unity in WebGL. We are currently unable to send a large amount of wearables. This is a mitigation of the error, while we implement proper pagination of wearables.

This issue only affects WebGL.

This PR removes duplicates from the wearable list, and checks the length of the resultant string. If its longer than the breaking cap, it adds only 600 wearables to the list and sends it with a warning. It adds 600 since we assume a bad length scenario (2000 chars) for every possible entry. 

# Why?


menduz (approved):
It looks good, thanks for commenting the code.

Didn't test it myself because I have not so much wearables, feel free to merge if you tested it

PR #780, opened 2022-12-14
feat: enable xmas empty scenes

leanmendoza:
# What? 
image

# Why? 
...


static/loader/empty-scenes/mappings.json
```diff
@@ -256,7 +328,15 @@
     },
     {
       "file": "models/EP_10.glb",
-      "hash": "QmTsvnJp8R2yrdADusiMXHdHjNe7PDr4SaPRd9Y6S1zLku"
+      "hash": "QmXmas210"
```
menduz, line 331: what, this doesn't look good? that's no hash

leanmendoza: I also wonder. Those scenes are already compiled

nearnshaw (approved):

PR #789, opened 2022-12-20
feat: add anti spam validations on friends requests feat

### LINKED ISSUE #698: Prevent friend requests from players that are blocked
- [ ] Check in own user profile that user is not blocked before processing the friend request
- [ ] Autoreject if the requesting user is currently blocked
- [ ] Add tests for the different use cases

### LINKED ISSUE #699: Set a cooldown timer between friend requests
- [ ] Implement in-memory check between friend requests on the sender side
- [ ] (TBD) Set a feature flag/explorer.json config for the configuration of the cooldown
- [ ] Add tests

### LINKED ISSUE #700: Prevent sending the same friend request more than N times
- [ ] Keep state of the number of friend requests sent in a session (in-memory) per requested user
- [ ] Avoid sending a friend request to the server if the maximum number was reached, otherwise forward the request
- [ ] Send spamming error to renderer when maximum number was reached
- [ ] Set a feature flag/explorer.json config for the maximum number of friend requests (default to 5)
- [ ] Add tests scenarios for success and error

Julieta11:
# What? 

So far, the PR addresses:
- Cooldown timer - Closes #699
- More than N times - Closes #700 
- Blocked users - Closes #698


packages/shared/friends/sagas.ts
```diff
@@ -622,14 +627,23 @@ function* refreshFriends() {
     const lastStatusOfFriends: Map<string, CurrentUserStatus> = yield getLastStatusOfFriends(store.getState()) ??
       new Map()
 
+    // Check if numberOfFriendRequests has values. If so, we keep them. If not, we initialize an empty map.
+    const numberOfFriendRequests: Map<string, number> = yield getNumberOfFriendRequests(store.getState()) ?? new Map()
+
+    // Check if coolDownOfFriendRequests has values. If so, we keep them. If not, we initialize an empty map.
+    const coolDownOfFriendRequests: Map<string, number> = yield getCoolDownOfFriendRequests(store.getState()) ??
+      new Map()
```
moliva, line 635: I think these should all take the more idiomatic saga form of:
```typescript
const state: Map<string, Something> = yield select(selectorFunction) ?? new Map()
```

packages/shared/friends/utils.ts
```diff
@@ -132,3 +133,23 @@ export function validateFriendRequestId(friendRequestId: string, ownId: string)
...
+  // Check whether the user is blocked
+  if (profile?.blocked?.includes(userId)) {
+    return true
+  }
+  return false
+}
+
+export const DEFAULT_MAX_NUMBER_OF_REQUESTS = 5
+
+export const COOLDOWN_TIME = 1000 // 1 second
```
moliva, line 155: Make constant name imply time unit.
```suggestion
export const COOLDOWN_TIME_MS = 1000 // 1 second
```

packages/shared/friends/sagas.ts
```diff
@@ -1425,6 +1444,11 @@ function* handleUpdateFriendship({ payload, meta }: UpdateFriendship) {
       yield call(refreshFriends)
     }
 
+    // Updates the friendship status of a blocked user to rejected after processing their incoming friend request.
+    if (FriendshipAction.REQUESTED_FROM === action && isBlocked(userId) && newFriendRequestFlow && incoming) {
+      yield call(handleBlockedUser, userId)
```
moliva, line 1449: Should this also include the request message since it's the new flow?

packages/shared/friends/sagas.ts
```diff
@@ -2487,3 +2523,78 @@ function buildFriendRequestErrorResponse(error: FriendshipErrorCode) {
...
+
+  // For each blocked user, update their friendship status as rejected
+  blockedIds.map(
+    async (fromFriendRequest) =>
+      await UpdateFriendshipAsPromise(
+        FriendshipAction.REJECTED,
+        fromFriendRequest.userId.toLowerCase(),
+        false,
+        fromFriendRequest.message
+      )
```
moliva, line 2589: Can we reuse the below `handleBlockedUser` here?

packages/shared/friends/utils.ts
```diff
@@ -132,3 +133,23 @@ export function validateFriendRequestId(friendRequestId: string, ownId: string)
...
+ */
+export function isBlocked(userId: string) {
+  // Get the user's profile
+  const profile = getCurrentUserProfile(store.getState())
+
+  // Check whether the user is blocked
+  if (profile?.blocked?.includes(userId)) {
+    return true
+  }
+  return false
```
moliva, line 150: Avoid the if and return the boolean expression (or false otherwise).
```suggestion
  return profile?.blocked?.includes(userId) ?? false
```

packages/shared/friends/sagas.ts
```diff
@@ -2487,3 +2523,78 @@ function buildFriendRequestErrorResponse(error: FriendshipErrorCode) {
...
+  )
+}
```
moliva, line 2591: Let's return a `Promise#all` of the above created promises so that the client of this function can actually await on it if need be!

packages/shared/friends/sagas.ts
```diff
@@ -2487,3 +2525,85 @@ function buildFriendRequestErrorResponse(error: FriendshipErrorCode) {
...
+ * @param fromFriendRequests - an array of from friend requests.
+ */
+function handleBlockedUsers(fromFriendRequests: FriendRequest[]) {
+  // Get the ids of users who have been blocked
+  const blockedIds = fromFriendRequests.filter((fromFriendRequest) => isBlocked(fromFriendRequest.userId))
+
+  // For each blocked user, update their friendship status as rejected
+  const promises = blockedIds.map(async (fromFriendRequest) => {
+    return handleBlockedUser(fromFriendRequest.userId, fromFriendRequest.message)
+  })
```
moliva, line 2573: Simplifying syntax.
```suggestion
  const promises = blockedIds.map(async (fromFriendRequest) =>
    handleBlockedUser(fromFriendRequest.userId, fromFriendRequest.message)
  )
```

packages/shared/friends/sagas.ts
```diff
@@ -2487,3 +2525,85 @@ function buildFriendRequestErrorResponse(error: FriendshipErrorCode) {
...
+  return Promise.all(promises)
+}
+
+/**
+ * Processes (update their friendship status as rejected) a friend request from a blocked user.
+ * @param id - the id of the user whose friend request is being processed.
+ */
+async function handleBlockedUser(id: string, message?: string) {
+  // Update their friendship status as rejected
+  await UpdateFriendshipAsPromise(FriendshipAction.REJECTED, id.toLowerCase(), false, message)
```
moliva, line 2585: Return promise result.
```suggestion
  return await UpdateFriendshipAsPromise(FriendshipAction.REJECTED, id.toLowerCase(), false, message)
```

packages/shared/friends/sagas.ts
```diff
@@ -757,6 +770,9 @@ export async function getFriendRequestsProtocol(request: GetFriendRequestsPayloa
     // Get friends
     const friends: FriendsState = getPrivateMessaging(store.getState())
 
+    // Reject blockedUsers
+    await handleBlockedUsers(friends.fromFriendRequests)
```
moliva, line 774: Log warning for users that could not be rejected in the background (below suggestion makes the return data available here).

packages/shared/friends/sagas.ts
```diff
@@ -2487,3 +2525,85 @@ function buildFriendRequestErrorResponse(error: FriendshipErrorCode) {
...
+function reachedMaxNumberOfRequests(userId: string) {
+  // Get number friend requests sent in the current session to the given user
+  const sentRequests = getNumberOfFriendRequests(store.getState())
+  const number = sentRequests.get(userId) ?? 0
+
+  // Get the maximum number of requests allowed
+  const maxNumber = DEFAULT_MAX_NUMBER_OF_REQUESTS
+
+  // Check if the current number of requests is less than the maximum allowed
+  return number < maxNumber ?? false
```
moliva, line 2543: This seems to work opposite [to the use here](https://github.com/decentraland/kernel/pull/789/files#diff-bcea93a768b74b3a44000978a9f8e8f2b257395933ac042653b8153f0a1c751fR2220).

Also, the default expression `?? false` should not be necessary here as number and maxNumber must always be defined if I understand.

Julieta11: Haha, right 🏃🏼‍♀️. I shouldn't have made this [commit](https://github.com/decentraland/kernel/pull/789/commits/664ea294bd1ab4e36c3198c1bb62eef270f89b74). It seems my brain got confused 🥴

packages/shared/friends/sagas.ts
```diff
@@ -757,6 +770,14 @@ export async function getFriendRequestsProtocol(request: GetFriendRequestsPayloa
     // Get friends
     const friends: FriendsState = getPrivateMessaging(store.getState())
 
+    // Reject blocked users
+    const blockedUsers = await handleBlockedUsers(friends.fromFriendRequests)
+    blockedUsers
+      .filter((blockedUser) => blockedUser.error !== null)
```
moliva, line 776: I know this is currently `null` when errors are reported, but in any case a more generic check might be better here (both `null` and `undefined` would be catched).
```suggestion
      .filter((blockedUser) => blockedUser.error)
```

moliva (commented):
Just a quick thing there and missing tests, but in general looks good to me! Great work here! 🦾 ✨

kuruk-mm (approved):
Looks great! Thanks!

PR #801, opened 2022-12-26
feat: teleport to home from world-realm

popuz:
# What? 
Closes #3780

# Why? 
After changing realm from world-realm to the normal-realm, position from world-realm was used, so user teleported to some place, that doesn't have any context for him

# Test steps
Scenario 1: 
**Given** you are in the `mgoldman.dcl.eth` world-realm
**When** changing realm via `/changerealm` chat message or changerealm button in UI to the any normal realm (loki for example)
**Then** you spawned at the 0,0 location in the target realm

Scenario 2: 
**Given** you are in the `mgoldman.dcl.eth` world-realm
**When** changing realm via `/goto` chat message or navmap JumpIn ui-button or Places JumpIn ui-button
**Then** you spawned at the chosen location/place


kwspgoliasz (approved):
The feature appears to be implemented correctly. The scenarios were tested in the mgoldman.dcl.eth world using the provided Decentraland link. Realm changing was functioning in accordance with the scenarios described in the ticket.

menduz (approved):

PR #802, opened 2022-12-28
Add Lambda 2 Public Url

mikhail-dcl:
Needed for unity-renderer/3834:
We are moving consuption of new Lambda endpoints to Renderer directly.

Once both PRs for Renderer and Kernel are approved I will move the tip of `protocol` to `dcl@next`

### DIFF
--- a/packages/shared/dao/index.ts
+++ b/packages/shared/dao/index.ts
@@ -185,7 +185,8 @@ async function resolveOfflineRealmAboutFromConnectionString(
         healthy: true,
         lambdas: {
           healthy: true,
-          publicUrl: `${baseUrl}lambdas`
+          publicUrl: `${baseUrl}lambdas`,
+          publicUrl2: `${baseUrl}lamb2`
         },
         acceptingUsers: true
       },
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "@dcl/hashing": "^1.1.2",
         "@dcl/kernel-interface": "^2.0.0-20210922153939.commit-017905d",
         "@dcl/legacy-ecs": "^6.11.8",
-        "@dcl/protocol": "^1.0.0-3766287015.commit-7d88411",
+        "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-3793446104.commit-ef6bc17.tgz",
         "@dcl/rpc": "^1.1.1",
         "@dcl/scene-runtime": "^7.0.3-20221223152034.commit-23102ad",
         "@dcl/schemas": "^5.29.0",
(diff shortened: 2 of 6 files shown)


packages/shared/dao/index.ts
```diff
@@ -185,7 +185,8 @@ async function resolveOfflineRealmAboutFromConnectionString(
         healthy: true,
         lambdas: {
           healthy: true,
-          publicUrl: `${baseUrl}lambdas`
+          publicUrl: `${baseUrl}lambdas`,
+          publicUrl2: `${baseUrl}lamb2`
```
menduz, line 189: This is not supposed to happen, instead lamb2 should be accesible via nginx re-route to lambdas @manumena please confirm

menduz (changes requested):

packages/shared/dao/index.ts
mikhail-dcl: I will be happy to remove the whole PR if it works this way.

mikhail-dcl: How can it work this way considering that the names of endpoints are the same (e.g. "profiles") for old and new API, and we will go through the feature flag at first?

manumena: @menduz @mikhail-dcl  lamb2 doesn't serve 3rd party wearables and can't be filtered by collectionId yet. The plan was
1. Focused to support the extended passport scope (pagination and other features)
2. Add 3rd party, collectionId and whatever is necessary to reach current functionality
3. Re-route lambdas to lamb2, keeping both endpoints available
4. Deprecate lambdas

We haven't started step 2 yet

manumena: @menduz 
the paths today would be like
- `/lambdas/wearables-by-owner` hit lambdas
- `/lamb2/nfts/wearables` hit lamb2

If your concern is to maintain the same service name for clarity reasons, we can do the re-routing right now and leave the endpoints like 
- `/lambdas/wearables-by-owner` hit lambdas
- `/lambdas/nfts/wearables` hit lamb2

and in another iteration

- `/lambdas/wearables-by-owner` hit lamb2
- `/lambdas/nfts/wearables` hit lamb2

so the name `lamb2` is not exposed

menduz: I'd like to not add extra configuration settings to all projects and move forward with the route of lamdas pointing to lamb2 to be consistent with the lamb2 migration plan

mikhail-dcl: @manumena @pentreathm will you have time to work on routing shortly? At the moment it will block us to proceed further if we want to stick with a single path.

pentreathm: yes, it's a small change. 
@manumena all yours :)

manumena: I'll start immediately

mikhail-dcl:
There will be no dedicated `lambd2` endpoint

PR #803, opened 2022-12-28
fix: add request dispatcher to friendRequestId

Julieta11:
For the new flow of Friend Request, we've added a field called `friendRequestId`, which is generated by kernel when a request is created. This field is used to manage the actions that can be executed on a friendship. Since Matrix doesn't provide a way to generate a `friendshipId`, we've decided to generate one by using the addresses of the users involved, with the rule and pattern `addressUser1` < `addressUser2` ?  `addressUser1_addressUser2` : `addressUser2_addressUser1`.

image

But there's a problem with that, because (see the [video](https://github.com/decentraland/unity-renderer/pull/3823#pullrequestreview-1230652236) to understand better) in that way, from renderer they cannot differentiate between friend requests and subsequent actions and they end up having different information for the same friendRequestId. They need to be able to differentiate between a request sent by User A to User B and a request sent by User B to User A.

**Note: we don't persist the friendRequestId. Therefore, it always needs to be derivable.**

**Solution I came up with:** 🫠
The rule is: `ownId` < `otherUserId` ? `ownId_otherUserId_requester` : `otherUserId_ownId_requester`

`requester` is the last 4 characters of the user id of the user who initiated the friendship, that is, sent the friend request. If the friend request is incoming and the action is either CANCELED or REQUESTED_FROM or if the friend request is outgoing and the action is either APPROVED or REJECTED, set the requester to the last 4 characters of the otherUserId. Otherwise set the requester to the last 4 characters of the ownId. In this way, we'll always be able to differentiate between a request sent by User A to User B and a request sent by User B to User A and the and subsequent actions that can be executed on a friendship.

### DIFF
--- a/packages/shared/friends/utils.ts
+++ b/packages/shared/friends/utils.ts
@@ -1,7 +1,7 @@
 import { getFeatureFlagEnabled, getFeatureFlagVariantValue } from 'shared/meta/selectors'
 import { RootMetaState } from 'shared/meta/types'
 import { store } from 'shared/store/isolatedStore'
-import { UsersAllowed } from 'shared/types'
+import { FriendshipAction, UsersAllowed } from 'shared/types'
 
 /**
  * Get the local part of the userId from matrixUserId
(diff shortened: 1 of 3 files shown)


packages/shared/friends/utils.ts
```diff
@@ -87,16 +87,40 @@ export function isNewFriendRequestEnabled(): boolean {
...
  */
-export function encodeFriendRequestId(ownId: string, otherUserId: string) {
+export function encodeFriendRequestId(ownId: string, otherUserId: string, incoming: boolean, action: FriendshipAction) {
   // We always want the friendRequestId to be formed with the pattern '0x1111ada11111'
   ownId = getUserIdFromMatrix(ownId)
   otherUserId = getUserIdFromMatrix(otherUserId)
 
-  return ownId < otherUserId ? `${ownId}_${otherUserId}` : `${otherUserId}_${ownId}`
+  // Dispatcher is the last 4 characters of the user id of the user who initiated the friendship, that is, sent the friend request
+  let dispatcher = ownId.substring(ownId.length - 4)
```
Julieta11, line 103: We set the dispatcher to the last 4 characters of the ownId in all scenarios except the one considered below

PR #837, opened 2023-01-10
fix: send friendshipStatus when updateUserPresence

Julieta11:
# What? 
Explorer expects the friendshipStatus attribute to be included when calling the UpdateUserPresence message, as it prevents inconsistencies when updating user presence.


lorux0 (approved):
great!

lorux0 (changes requested):
no need for this changes, renderer is ignoring the friendship status in the presence message

Julieta11:
Closing this pull request as we have realized that Explorer does not expect the friendshipStatus attribute to be included when calling the UpdateUserPresence message

PR #847, opened 2023-01-16
fix: check last sent profile status

Julieta11:
We are sending a lot of "passport" messages (`AddUserProfileToCatalog`) from the kernel to the renderer, even when the profile doesn't have any updates. So, in order to avoid that, we just check that the version is greater than the previous one sent to renderer.


packages/shared/renderer/sagas.ts
```diff
@@ -295,11 +296,19 @@ function* handleSubmitProfileToRenderer(action: SendProfileToRenderer): any {
...
       baseUrl: fetchContentServerWithPrefix
     })
-    getUnityInstance().AddUserProfileToCatalog(forRenderer)
-    yield put(addedProfileToCatalog(userId, profile.data))
+
+    // Add deep equal check before submitting profile to renderer
+    if (!lastSentProfileStatus || !deepEqual(lastSentProfileStatus, forRenderer)) {
+      getUnityInstance().AddUserProfileToCatalog(forRenderer)
+      yield put(addedProfileToCatalog(userId, profile.data))
+      yield put(addProfileToLastSentProfileStatuses(userId, forRenderer))
```
moliva, line 310: We should merge the two of these into a single action to avoid inconsistencies.

packages/shared/renderer/sagas.ts
```diff
@@ -295,11 +296,19 @@ function* handleSubmitProfileToRenderer(action: SendProfileToRenderer): any {
...
+    const lastSentProfileStatus = lastSentStatuses[userId]
+
     const forRenderer = profileToRendererFormat(profile.data, {
       baseUrl: fetchContentServerWithPrefix
     })
-    getUnityInstance().AddUserProfileToCatalog(forRenderer)
-    yield put(addedProfileToCatalog(userId, profile.data))
+
+    // Add deep equal check before submitting profile to renderer
+    if (!lastSentProfileStatus || !deepEqual(lastSentProfileStatus, forRenderer)) {
```
moliva, line 307: I would avoid the `deepEqual` here in favor of checking that the `version` of the new profile is great than the one already in store.

moliva (approved):
LGTM!

test/unit/profiles.saga.test.tsx
```diff
@@ -64,22 +77,21 @@ describe('fetchProfile behavior', () => {
...
-        .dispatch(profileRequest('user|1', future()))
-        .provide([
-          [select(getRealmAdapter), {}],
-          [call(profileServerRequest, 'user|1'), delayedProfile],
-          [select(getCurrentUserId), 'myid'],
-          [call(ensureAvatarCompatibilityFormat, profile), 'passport']
-        ])
-        .run()
-    })
+  it.skip('completes once for more than one request of same user', () => {
```
eordano, line 80: Why is this test skipped?

Julieta11: I'm not sure. They were like that from before. Maybe @pbosio can shed some light for us since from what I see, he was the one who committed the last time.

pbosio: after we migrated kernel to this repository, we find out that lots of test weren't running and, after making them run, lots of test were failing, so some test were fixed, others coded again and probably a couple skipped.
regarding this test in particular, I think the idea back then was to refactor the profile sagas in the short term, so we just skipped that test instead of fixing it.

packages/shared/renderer/sagas.ts
```diff
@@ -295,11 +294,18 @@ function* handleSubmitProfileToRenderer(action: SendProfileToRenderer): any {
...
   } else {
-    const forRenderer = profileToRendererFormat(profile.data, {
-      baseUrl: fetchContentServerWithPrefix
-    })
-    getUnityInstance().AddUserProfileToCatalog(forRenderer)
-    yield put(addedProfileToCatalog(userId, profile.data))
+    const lastSentProfileVersion: number | undefined = yield select(getLastSentProfileVersion, userId)
+
+    // Add version check before submitting profile to renderer
+    if (!lastSentProfileVersion || lastSentProfileVersion < profile.data.version) {
```
pbosio, line 300: I don't think it's possible right now because of our current flow where new users are forced to create an avatar at the very beginning, but technically profile version might be `0` and make `!lastSentProfileVersion` always `true`

pbosio (approved):

packages/shared/renderer/sagas.ts
Julieta11: Great comment! Thx! 🙌🏼

